PR #62, opened 2019-02-28
Hotfix for Preproduction deployment: Reports not rendering and rendering with wrong font

dy2288:


dy2288:
Cherry pick rather than merge.

PR #68, opened 2019-03-14
Fixing ProviderHelpers.js > getDialectPathFromURLArray()

coakenfold:
Includes jest test but this branch isn't currently setup to support jest tests (it's supported in the `jest-integration` branch).

FROM:
```
  getDialectPathFromURLArray: function(url) {
    return decodeURI(url.slice(1, 7).join('/'))
  },
```

TO:
```
/*
getDialectPathFromURLArray(urlArray)

Extracts dialect path from url array.
Returns null if not found
*/
function getDialectPathFromURLArray(urlArray) {
  const index = urlArray.findIndex((el) => {
    return el.match(/^FV$/)
  })
  if (index !== -1) {
    const _url = urlArray.slice(index, index + 5)
    return decodeURI(`/${_url.join('/')}`)
  }
  return null
}
```

Note: `ProviderHelpers.js` has been reorganized to have the fns() outside of the exported object

### DIFF
--- a/frontend/app/assets/javascripts/common/ProviderHelpers.js
+++ b/frontend/app/assets/javascripts/common/ProviderHelpers.js
@@ -16,23 +16,9 @@ limitations under the License.
 
 import Immutable, { List, Map } from 'immutable'
 import StringHelpers from './StringHelpers'
-
 import selectn from 'selectn'
 
-const toJSKeepId = function(js) {
-  return typeof js !== 'object' || js === null
-    ? js
-    : Array.isArray(js)
-      ? Immutable.Seq(js)
-        .map(toJSKeepId)
-        .toList()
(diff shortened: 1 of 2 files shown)


frontend/app/assets/javascripts/common/ProviderHelpers.js
```diff
@@ -16,23 +16,9 @@ limitations under the License.
...
-      ? Immutable.Seq(js)
-        .map(toJSKeepId)
-        .toList()
-      : js.hasOwnProperty('id')
-        ? Immutable.Seq(js).toMap()
-        : Immutable.Seq(js)
-          .map(toJSKeepId)
-          .toMap()
-}
-
```
coakenfold: Also rewrote this ternary to `if`s and `else if`s. See line #274

PR #77, opened 2019-04-02
Prettiered

coakenfold:
Added a `.prettierignore` and then ran `prettier --write "[some directory]/**/*.js"` in the terminal


coakenfold:
thanks!

PR #82, opened 2019-07-05
Create Word Form Part of speech optional

gmcauliffe:
- Updated Create WOrd form so that part of speech defaults to "Please Select:" and is an optional entry.
- Added a "placeholder" boolean prop to "Directory list", if true will add a placeholder to the field.
- Updated field.js email verification to pass  Husky

### DIFF
--- a/frontend/app/assets/javascripts/views/pages/explore/dialect/create/Word/validation.js
+++ b/frontend/app/assets/javascripts/views/pages/explore/dialect/create/Word/validation.js
@@ -5,10 +5,7 @@ import copy from './internationalization'
 
 const validForm = yup.object().shape({
   'dc:title': yup.string().required(copy.validation.title),
-  'fv-word:part_of_speech': yup
-    .string()
-    .label('Part of speech')
-    .required(),
+  'fv-word:part_of_speech': yup.string().label('Part of speech'),
   'fv-word:pronunciation': yup.string(),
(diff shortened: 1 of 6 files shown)


frontend/app/assets/javascripts/views/pages/explore/dialect/create/Word/validation.js
```diff
@@ -5,10 +5,7 @@ import copy from './internationalization'
 
 const validForm = yup.object().shape({
   'dc:title': yup.string().required(copy.validation.title),
-  'fv-word:part_of_speech': yup
-    .string()
-    .label('Part of speech')
-    .required(),
+  'fv-word:part_of_speech': yup.string().label('Part of speech'),
```
coakenfold: Could the string 'Part of speech' be added to the internationalization.js file @ frontend/app/assets/javascripts/views/pages/explore/dialect/create/Word/internationalization.js?

Like how line 7 is doing it above

gmcauliffe: Definitely, can do.
On review though, do we still need to use ".label" to override the key name if it is no longer a required field?

coakenfold: Yeah you're right, it's not needed on an optional field

coakenfold:
Thanks!

PR #83, opened 2019-07-16
FW-169 update phrase detail view

gmcauliffe:
Phrase detail view updated to match Word detail view and include audio details

https://firstvoices.atlassian.net/browse/FW-169?atlOrigin=eyJpIjoiY2RmN2E0ZWU4ZjMzNDYxZmE0YmNjZjAyOGRmMTJiYWYiLCJwIjoiaiJ9


frontend/app/assets/javascripts/views/pages/explore/dialect/learn/phrases/view.js
```diff
@@ -98,16 +94,31 @@ export class View extends Component {
...
+      '_getDefinitions',
+      // '_getPhotos',
+      // '_getPhrases',
+      // '_getPronounciation',
+      '_getTabs',
+      // '_getVideos',
+      '_getPhrasePath',
+      // '_groupBy',
+      '_onNavigateRequest',
+    ].forEach((method) => (this[method] = this[method].bind(this)))
```
coakenfold: If you convert the functions to use the fat arrow syntax all of this can be dropped (lines 104-116)

For example, this:
`_getPhrasePath(props = null) {`

Can be updated to this:
`_getPhrasePath = (props = null) => {`

Not a blocker for the PR, just an FYI. It'll help reduce the bundle sizes.

frontend/app/assets/javascripts/views/pages/explore/dialect/learn/phrases/view.js
```diff
@@ -268,60 +202,53 @@ export class DialectViewPhrase extends Component {
...
-                <div className="DialectViewWordTitleAudio">
-                  <h2 className={`DialectViewWordTitle ${dialectClassName}`}>
-                    {title} {this._getAudio(computePhrase)}
-                  </h2>
-                </div>
-                {this._getDefinitions(computePhrase)}
-                {this._getCulturalNotes(computePhrase)}
-                {/* {this._getLiteralTranslations(computePhrase)}
-                {this._getPronounciation(computePhrase, computeDialect2)} */}
+        <main className="DialectViewWord" id="contentMain">
```
coakenfold: Wondering about the use of `<main />`, I checked at MDN and it says:

> The content of a `<main>` element should be unique to the document. Content that is repeated across a set of documents or document sections such as sidebars, navigation links, copyright information, site logos, and search forms shouldn't be included unless the search form is the main function of the page.

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main

coakenfold:
Looks good to me, didn't see any issues.

frontend/app/assets/javascripts/views/pages/explore/dialect/learn/phrases/view.js
coakenfold: ^ Not saying this is incorrect, it's just not easy to tell from reading the code where the `<main />` shows up on the page.

frontend/app/assets/javascripts/views/pages/explore/dialect/learn/phrases/view.js
gmcauliffe: Great! Thanks for the tip, much appreciated.

frontend/app/assets/javascripts/views/pages/explore/dialect/learn/phrases/view.js
gmcauliffe: TBH I was just trying to maintain consistency and used `<main />` because it was used in `word/view.js`.

What would you suggest to make it clearer?

coakenfold: (this is a really minor issue and doesn't have strong repercussions, but..)

I'd just check to see where the code shows up on the page. If it's in a sidebar and/or there is already a `<main />` in the dom, then it should probably be changed to a different tag.

The MDN docs say:
> A document mustn't have more than one `<main>` element that doesn't have the hidden attribute specified.

So that makes me think `<main />` should be reserved for the main content of the page. 

That could mean only one `<main />` across the entire site, perhaps updating the `<div id="pageContainer" className="AppFrontController__content">` in `AppFrontController.js` 

or

each 'page level' component defines a single `<main />`, so there would be multiple `<main />`s in the code base but only one `<main />` in the dom at a time, eg: `<main><h1>Create Word</h1><form>...</form></main>`

coakenfold: Oh and the two things I highlighted aren't blockers for merging the PR, more of a heads up. Feel free to merge without making any changes.

gmcauliffe: Thanks so much! Much appreciated.

PR #85, opened 2019-07-23
FW-249 Words Phrases Broken, Cypress tests

coakenfold:
https://firstvoices.atlassian.net/browse/FW-249
Bug caused by 2 issues:
* Number regex incorrectly matching UIDs in url because it was missing start quantifier: ^ (already had end quantifier in place: $). So now the updated regex will match a string if it starts, ends, and is entirely made up of numbers.
* `NavigationHelpers.navigateForwardReplaceMultiple` called on urls without pagination segments. Resulted in urls like `.../dialect/learn/words` getting updated to `.../dialect/10/1` (which caused 404s)

https://firstvoices.atlassian.net/browse/FW-250
Added Cypress tests for major features related to: Search (Phrase & Word, Private & Public), Alphabet, Category, & Flashcards


frontend/app/assets/javascripts/views/components/AlphabetListView/__cypress__/AlphabetListView-Phrases.js
```diff
@@ -0,0 +1,76 @@
...
+    cy.visit('http://0.0.0.0:3001/nuxeo/app/explore/FV/sections/Data/Athabascan/Dene/Dene/learn/phrases')
+
+    const letter = 'b'
+    const unselectedColor = 'rgb(60, 52, 52)'
+
+    // No message, button, or selected letters
+    cy.log('No message, button, or selected letters')
+    cy.queryByText(/showing phrases that start with the letter/i).should('not.exist')
+    cy.queryByText(/stop browsing alphabetically/i).should('not.exist')
+    cy.getByTestId('AlphabetListView').within(() => {
```
coakenfold: `cy.getByTestId('AlphabetListView')` is an escape hatch for selecting parts of the interface. It looks for `data-testid="AlphabetListView"` in the dom. 

We resort to using `data-testid` properties when there are similar UI elements on the page and could receive an unintended element.

For example, it's possible a page could have several buttons that are labeled 'Search', if we call `cy.getByLabelText('Search')` we can't be certain which button is selected.

Another use case is to limit the focus to a section of the UI and then search generic tags. Like what the code attached to this comment is doing. It's looking in the parent dom element (with `data-testid="AlphabetListView"`) and loops through all `<a />`s within it.

frontend/app/assets/javascripts/views/components/AlphabetListView/__cypress__/AlphabetListView-Phrases.js
```diff
@@ -0,0 +1,76 @@
...
+      cy.get('a').each(($el) => {
+        cy.wrap($el)
+          .should('have.css', 'color')
+          .and('eq', unselectedColor)
+      })
+    })
+    cy.log('Test complete')
```
coakenfold: These are useful to show up in the Cypress runner. Sometimes you won't be certain if the test is finished or hung up on the last step.

frontend/app/assets/javascripts/views/components/Browsing/dictionary-list.js
```diff
@@ -92,6 +92,7 @@ export default class DictionaryList extends Component {
           {(items || []).map((item, i) => (
             <tr
               key={i}
+              data-testid="DictionaryList__row"
```
coakenfold: Example of adding a `data-testid`

frontend/app/assets/javascripts/views/components/SearchDialect/__cypress__/SearchDialect-Phrases-Private.js
```diff
@@ -0,0 +1,33 @@
+// NOTE: this file will be copied to `cypress/integration` and run from there,
+// so imports paths will be based on that location!
+import 'cypress-testing-library/add-commands'
+import testSearchPhrases from '../../../app/assets/javascripts/views/components/SearchDialect/__cypress__/common/testSearchPhrases.js'
+
+describe('SearchDialectPhrasesPrivate', () => {
+  it('Should redirect with anon user, no redirect with member', () => {
+    cy.log('Trying to access private section with anon user')
+    cy.visit('http://0.0.0.0:3001/nuxeo/app/explore/FV/Workspaces/Data/Athabascan/Dene/Dene/learn/phrases')
+    cy.wait(500)
```
coakenfold: These waits are useful to add to tests. Sometimes Cypress will confirm something before the browser has a chance to change urls or Cypress will have a hard time finding an element and will trigger an error regarding `Maximum call stack exceeded`.

I've found adding a `cy.wait` in those scenarios leads to more reliable tests.

frontend/app/assets/javascripts/views/components/SearchDialect/__cypress__/SearchDialect-Phrases-Private.js
```diff
@@ -0,0 +1,33 @@
...
+    cy.location('pathname').should('eq', '/nuxeo/app/explore/FV/sections/Data/Athabascan/Dene/Dene/learn/phrases')
+    cy.log('Trying to access private section with registered user')
+    cy.login({
+      url: 'https://preprod.firstvoices.com/nuxeo/startup',
+      data: {
+        user_name: Cypress.env('ADMIN_USERNAME'),
+        user_password: Cypress.env('ADMIN_USERNAME'),
```
coakenfold: Setting Login/Pass via environment variables on my computer:
https://docs.cypress.io/guides/guides/environment-variables.html#Option-3-CYPRESS

I've set the vars in my`.bash_profile`:
`export CYPRESS_ADMIN_USERNAME='THE-USERNAME'`
`export CYPRESS_ADMIN_PASSWORD='THE-PASSWORD'`

frontend/package.json
```diff
@@ -7,7 +7,7 @@
   "scripts": {
     "build": "gulp build:production",
     "build:production": "npm run production; npm run production:legacy",
-    "cy:copy": "npx copyfiles ./app/assets/javascripts/**/__cypress__/*.js cypress/integration/__cypress__ -f",
+    "cy:copy": "npx copyfiles -e ./app/assets/javascripts/**/__cypress__/common/*.js ./app/assets/javascripts/**/__cypress__/*.js cypress/integration/__cypress__ -f",
```
coakenfold: Copies any files it finds within `__cypress__` directories to `cypress/integration/__cypress__` but excludes any files found in `__cypress__/common`.

The `common` dir is intended for cypress related functions that are shared between similar tests. An example can be found in `.../SearchDialect/__cypress__/`

PR #102, opened 2019-10-21
Match v0 styling more closely

coakenfold:
Includes fixes for:

FW-620 [v1] Style Navigation more closely to v0
FW-617 [MAT-UI] Define typography to match v0
FW-616 [MAT-UI] Buttons are styled using primary/secondary theme properties but we need customized colours outside of primary/secondary


coakenfold:
Screen Shot 2019-11-01 at 10 09 56 AM

dy2288:
Looks good @coakenfold

PR #121, opened 2019-12-03
FW-698 Create FirstVoicesData listener to replace xml FVFillAncestors

NolanVH:
Adds a listener in FirstVoicesData to replace xml FVFillAncestors in ca.firstvoices.studio.extensions.xml to fix error 500 on local backend setup.


FirstVoicesData/src/main/resources/OSGI-INF/ca.firstvoices.listeners.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0"?>
+<component name="ca.firstvoices.listeners.FVDocumentListener" version="1.0">
+
+    <extension point="listener" target="org.nuxeo.ecm.core.event.EventServiceComponent">
+        <listener enabled="true" name="fvdocumentlistener" async="false" postCommit="false"
```
dy2288: Are we not able to ensure this event fires only for those document types on this point? (in XML)

NolanVH: I tried using the <filters> tag here, but when I tested it the listener did not seem to notice it at all and would fire on other documents as well.

dy2288:
Good resources to read @NolanVH 
- https://doc.nuxeo.com/studio/how-to-unit-test-a-studio-bundle/
- https://doc.nuxeo.com/corg/unit-testing/

PR #122, opened 2019-12-05
FW-691 Add functionality to create/delete TEst/Test directory automatically

NolanVH:
The database setup script will now create the TEst/Test directory inside of FV/Workspaces/Data/ if it doesn't exist.

The database teardown script will now delete the TEst/Test directory inside of FV/Workspaces/Data if there are no files or directories left inside of it after the teardown is complete.

Also added some echo lines to make the output more readable.


NolanVH:
> @NolanVH is there a reason why we are maintaining the typo TEst/Test (with the capital E?) Let's fix that.

I have added a commit to this pull request which fixes the typo. The directory created will now be Test/Test

PR #166, opened 2020-02-01
FW-677 Update Cypress tests with to run on CI and enable the GitHub Actions workflow

NolanVH:
This pull request updates all Cypress tests to run on GitHub Actions CI. All pull requests to master will trigger the tests to run. Pull requests will have a Cypress bot comment indicating which tests passed and failed.  Test runs can be viewed while running and after completion at [this link.](https://dashboard.cypress.io/projects/gdqzxg/runs) A large portion of the later commits are small changes (mainly adding waits) to the Cypress tests to get them passing on slower CI machines.


NolanVH:
The Cypress test "Navigation.js" is the only one failing currently, due to FW-840 (see comments in issue).

frontend/app/assets/javascripts/views/pages/explore/dialect/learn/words/__cypress__/RecorderPublishUnpublish-Word.js
```diff
@@ -111,19 +127,20 @@ describe('RecorderPublishUnpublish-Word.js > RecorderPublishUnpublish-Word', ()
       userName: 'TESTLANGUAGEFIVE_ADMIN',
     })
     cy.visit('/explore/FV/Workspaces/Data/Test/Test/TestLanguageFive/')
-    cy.wait(500)
+    cy.wait(1000)
     cy.getByText('View My Tasks', { exact: false }).click()
     cy.getByText('Reject', { exact: true }).click()
     cy.wait(500)
     cy.visit('/explore/FV/Workspaces/Data/Test/Test/TestLanguageFive/learn/words')
+    cy.wait(5000)
```
coakenfold: Are the 5 second `.wait()`s in this file intentional? I see some 500s and just wanted to check.

NolanVH: Yes. This test was particularly hard to get working on the slower CI machines and seemed to need more wait time when using cy.visit().

coakenfold: Ok, thanks for the confirmation.

NolanVH:
> Looks good to me. I don't recommend any changes that would block merging of this PR.
> 
> However there are a couple of things that may be useful to revisit in the future:
> 
> * See if it's possible to minimize the amount of `cy.wait()` calls. Cypress says it's not a best practice **but** the examples they provide are waiting for requests to finish and not an SPA displaying new content. https://docs.cypress.io/guides/references/best-practices.html#Unnecessary-Waiting
> * There are still a few selectors that are traversing up the DOM (via repeated calls to `.parent()`). These tests could break with minor changes to the DOM. If there are no good hooks at the target node, consider adding a `data-testid` attribute to the jsx

I agree with both of your points. It would be good to reduce number of `cy.wait()` calls used but with our current implementiation it may not be possible.  I do think tests in a few places could be fragile due to hooks used and adding more `data-testid` to the jsk would go a long way to helping this issue.

Thanks for the review Chad!

PR #178, opened 2020-02-11
FW-853 Cypress Tests > Phrase/Word > Clicking Browse

coakenfold:
Word/Phrase tests now start by:
- creating browse items with names that start with a timestamp
- clears out the form
- selects the new items via the browse components
- populates the rest as normal
- saves
- verifies the browse items are on the detail page
(continues the test as before)


coakenfold:
Thanks @NolanVH

PR #189, opened 2020-02-21
Update CI failing Cypress tests to wait for XHRs instead of fixed waits

NolanVH:


dy2288:

PR #197, opened 2020-02-28
FW-694 FW-796 Refactor InitialDatabaseSetup.java, add unit test

NolanVH:
Also adds a proper message and continues when no environent variables are set for the creation of an Admin user, instead of giving a 500 error.


dy2288:

PR #201, opened 2020-02-29
Fix CRLF to LF issue which would cause the transitions to fail on CI

NolanVH:


dy2288:

PR #202, opened 2020-03-02
FW-749 - Use lightweight tags for version

dy2288:


dy2288:

PR #206, opened 2020-03-02
Fix initialsetup.sh return code error on success

NolanVH:


dy2288:

PR #209, opened 2020-03-03
FW-1016 Add docker module deploy script

NolanVH:
The purpose of this change is to provide developers an easy way to deploy backend changes to their local dev environment quickly.

To use the script navigate to the docker folder and run ```./UpdateModule.sh <ModuleName>``` where <ModuleName> is the name of the new or changed module that you want to deploy to your local backend.  This will build the module, remove any old copies from inside of the docker container, copy the new jar file into the container, and reboot the nuxeo backend.


dy2288:

davidlacho:
Can the command be called from inside a project directory? So rather than navigating to the docker directory to deploy changes for a single module, could we have it so that each project can call the script that is found in the docker directory?

NolanVH:
> Can the command be called from inside a project directory? So rather than navigating to the docker directory to deploy changes for a single module, could we have it so that each project can call the script that is found in the docker directory?

Scripts have been added into each module which can be called from the root of that module to perform the deploy.  The only command needed is now ```./UpdateModule.sh``` from the root of the current module.

> Receiving error on running `./UpdateModule.sh FirstVoicesData`:
> 
> ```
> Removing old bundle from docker container if it exists.
> rm: cannot remove '/opt/nuxeo/server/nxserver/bundles/-*.jar': No such file or directory
> 
> Copying new bundle into docker container and restarting nuxeo backend.
> Nuxeo home:          /opt/nuxeo/server
> Nuxeo configuration: /etc/nuxeo/nuxeo.conf
> Include template: /opt/nuxeo/server/templates/common-base
> Include template: /opt/nuxeo/server/templates/common
> Include template: /opt/nuxeo/server/templates/default
> Include template: /opt/nuxeo/server/templates/docker
> Include template: /opt/nuxeo/server/templates/first-voices
> 
> Stopping server......Server stopped.
> Configuration files generation (nuxeo.force.generation=true)...
> Configuration files generated.
> Server start failed (2).
> ERROR: transport error 202: bind failed: Address already in use
> ERROR: JDWP Transport dt_socket failed to initialize, TRANSPORT_INIT(510)
> JDWP exit error AGENT_ERROR_TRANSPORT_INIT(197): No transports initialized [debugInit.c:750]
> ```

This error has been fixed and the changes have been added to this pull request.

Thanks for the review @davidlacho!

PR #210, opened 2020-03-04
FW-672 Automate call to initial setup on backend docker startup

NolanVH:


NolanVH:
> LGTM: Checked out the branch and was able to run docker/site without any errors. Seems like the initialsetup script was run since I could login with an admin account.

Thanks for the review @coakenfold!

PR #212, opened 2020-03-06
FW-1058

Dr-Slugma:
This pull request should make it so that as long as a language has enough valid words, a user can always play parachute and parachute will find words.


frontend/app/assets/javascripts/views/pages/explore/dialect/play/hangman/index.js
```diff
@@ -78,12 +79,35 @@ export class Hangman extends Component {
...
         '&pageSize=' +
         PUZZLES
     )
-
-    
+      
+      const queryResult = ProviderHelpers.getEntry(this.props.computeWords,
+        this.props.routeParams.dialect_path + '/Dictionary')
+
+      this.setState({resultCount: selectn(['response', 'resultsCount'], queryResult)})
```
coakenfold: Perhaps this `setState` is premature? One could see if the 2nd request (below, lines 92+) needs to run before setting state. 

It may save a re-render and could possibly sidestep issues around batch processing and the async nature of `setState` updates: https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous

PR #216, opened 2020-03-09
release-3.1.4-HF01

coakenfold:
Addresses:
FW-1013
FW-1042
FW-1063


dy2288:
For a hot-fix we actually don't want a pull request (since we'll be releasing from the branch directly)

PR #217, opened 2020-03-10
FW-1006: Refactor fv:custom-order generation algorithm

davidlacho:
Jira ticket:
https://firstvoices.atlassian.net/browse/FW-1006

This ticket handles checking graphemes against a word such that it will always decompose a word into the proper grapheme, handles multi-character representation and Unicode representation of diacritics, etc, both capitalized and lowercase form of graphemes are handled so that there is no difference in custom order, and spaces and characters outside of the alphabet are pushed to the end of the sort order.


FirstVoicesPublisher/src/main/java/ca/firstvoices/nativeorder/services/NativeOrderComputeServiceImpl.java
```diff
@@ -9,123 +18,136 @@
...
+                            ":alphabet_order"))).toString();
                     title = title.substring(charValue.length());
                     found = true;
                     break;
                 }
             }
             if (!found) {
                 if (" ".equals(title.substring(0, 1))) {
-                    nativeTitle += new Character((char) (33 + chars.length + 1)).toString();
+                    nativeTitle += "~";
```
Dr-Slugma: I think it might be a good idea to follow the example of [the Oxford Dictionary](https://www.oxfordlearnersdictionaries.com/wordlists/oxford-phrase-list) and have spaces come _before_ letters in the sort order.  (Therefore spaces use "!" and special characters use "~")

Also, we should ensure that all alphabet orders start at 1, not 0.

PR #220, opened 2020-03-13
FW 1007 remove spaces

Dr-Slugma:
This pull request involves the creation of a service to remove spaces before and after the title of a word and/or phrase.


FirstVoicesData/src/test/java/ca/firstvoices/testUtil/SanitizeDocumentTestUtil.java
```diff
@@ -0,0 +1,85 @@
+package ca.firstvoices.testUtil;
+
+import org.nuxeo.ecm.core.api.*;
+import org.nuxeo.ecm.core.api.DocumentModel;
+import org.nuxeo.ecm.core.api.CoreSession;
+import static org.junit.Assert.assertNotNull;
+
+public class SanitizeDocumentTestUtil{
```
davidlacho: Since this is repeating a lot of the same code in the other test util, and if you could see these methods being used in future tests, I would recommend abstracting both classes and any of our test cases can extend the AbstractTest.

FirstVoicesData/src/main/java/ca/firstvoices/listeners/FVDocumentListener.java
```diff
@@ -42,8 +46,11 @@ public void handleEvent(Event event) {
     CoreSession session = ctx.getCoreSession();
     
     // If the document is mutable and is the proper type then assign its ancestors.
-    service.assignAncestors(session, document);
-    
+    if(event.getName().equals(DocumentEventTypes.DOCUMENT_CREATED))
```
davidlacho: Since your service only applies to FVWord and FVPhrase, we should discuss restructuring these packages to be related to the specific type. FVDocument should apply to _all_ FVDocuments.

FirstVoicesData/src/main/java/ca/firstvoices/services/SanitizeDocumentServiceImpl.java
```diff
@@ -0,0 +1,29 @@
...
+
+import org.nuxeo.ecm.core.api.CoreSession;
+import org.nuxeo.ecm.core.api.DocumentModel;
+
+public class SanitizeDocumentServiceImpl implements SanitizeDocumentService {
+  
+    //Sanitizes the document via various functions, and saves the document if edited
+    public void sanitizeDocument(CoreSession session, DocumentModel currentDoc) {
+        boolean edited = false;
+        if((currentDoc.getType().equals("FVWord") || currentDoc.getType().equals("FVPhrase"))
```
davidlacho: Since your service only applies to FVWord and FVPhrase, we should discuss restructuring these packages to be related to the specific type. FVDocument should apply to all FVDocuments. Alternatively, if you foresee this service being used for other FVDocuments leave it in that module.

What's your plan for this service?

FirstVoicesData/src/main/java/ca/firstvoices/services/SanitizeDocumentServiceImpl.java
```diff
@@ -0,0 +1,29 @@
...
+        
+        if(edited) session.saveDocument(currentDoc);
+    }
+
+    //Trims spaces if any spaces need to be trimmed
+    private Boolean trimSpaces(DocumentModel currentDoc){
+
+        String title = currentDoc.getTitle();
+        if(!title.equals(title.trim())){        
+            currentDoc.setPropertyValue("dc:title", title.trim());
```
davidlacho: As above, if this service will be used for trimming specific properties outside of "dc:title", do you have a way to account for that? Say, for example, I wanted to trim whitespace from a different property?

FirstVoicesData/src/main/java/ca/firstvoices/services/SanitizeDocumentServiceImpl.java
```diff
@@ -0,0 +1,29 @@
...
+            return true;
+        }
+        return false;
+    }
+}
```
davidlacho: Considering the facts above, I would refactor this service method to be re-usable for different document types, and different properties.

FirstVoicesData/src/main/java/ca/firstvoices/services/SanitizeDocumentServiceImpl.java
```diff
@@ -0,0 +1,29 @@
+package ca.firstvoices.services;
+
+import org.nuxeo.ecm.core.api.CoreSession;
+import org.nuxeo.ecm.core.api.DocumentModel;
+
+public class SanitizeDocumentServiceImpl implements SanitizeDocumentService {
+  
+    //Sanitizes the document via various functions, and saves the document if edited
+    public void sanitizeDocument(CoreSession session, DocumentModel currentDoc) {
```
davidlacho: Sanitize document is not a great method name. A better name might be better here such as "remove whitespace". I would also pass in the name of the property field that you would like to remove whitespace from for re-usability.

FirstVoicesData/src/test/java/ca/firstvoices/services/SanitizeDocumentServiceImplTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+    assertEquals(TestWord.getTitle(), " Test Word ");
+    assertEquals(TestPhrase.getTitle(), "  Test Phrase  ");
+
+    // Run the service against the word & phrase documents
+    sanitizeDocumentServiceInstance.sanitizeDocument(session, TestWord);
+    sanitizeDocumentServiceInstance.sanitizeDocument(session, TestPhrase);
+
+    assertEquals(TestWord.getTitle(), "Test Word");
+    assertEquals(TestPhrase.getTitle(), "Test Phrase");
```
davidlacho: We should also be testing our event listeners. Does your method get fired when the document is saved and modified?

FirstVoicesData/src/main/java/ca/firstvoices/services/SanitizeDocumentServiceImpl.java
Dr-Slugma: I was thinking that in the future we could extend this service to perhaps do tasks such as composite characters -> unicode.  As well, I think that space removal for titles is probably something we would want for all documents, but the ticket focuses on words & phrases.

FirstVoicesData/src/main/java/ca/firstvoices/services/SanitizeDocumentServiceImpl.java
Dr-Slugma: I currently do not, but I could have the method take in a property as a parameter.

PR #221, opened 2020-03-13
FW-935 FW-1046 Create unpublished changes service, endpoint operation, enricher

NolanVH:
These changes create a service which can be used to check if there are unpublished changes on a document. In addition to the service an endpoint and enricher are provided.

The endpoint is called FVCheckUnpublishedChanges and returns a boolean true if unpublished changes exists. The endpoint calls the service to do the check.

The enricher calls the service and adds a json object which has a key "unpublished_changes_exist" and a true value if the document has unpublished changes, otherwise false.  The enricher could be further developed in the future to return information detailing what items within the document are unpublished.  To use the enricher simple add a call to it in a frontend header such as [this line](https://github.com/First-Peoples-Cultural-Council/fv-web-ui/blob/6609db692c61261b900d780623058266115ced93/frontend/app/assets/javascripts/providers/redux/reducers/fvWord/actions.js#L21) and inspect the fetch response. 
eg. change the line to 
> 'enrichers.document': 'ancestry,word,permissions,unpublished_changes',


FirstVoicesEnrichers/src/test/java/ca/firstvoices/UnpublishedChangesEnricherTest.java
```diff
@@ -0,0 +1,145 @@
...
+        json = json.has("contextParameters").isObject();
+        json.properties(1);
+        json = json.has(UnpublishedChangesEnricher.NAME).isObject();
+        json.has("unpublished_changes_exist").isEquals(false);
+    }
+
+    /*
+        Helper method to create a dialect tree.
+     */
+    protected void createDialectTree() throws Exception {
```
davidlacho: These helper methods should be put into an abstract class that all tests inherit from. These are handy methods and we're starting to see them duplicated around the codebase often enough.

FirstVoicesPublisher/src/main/java/ca/firstvoices/services/UnpublishedChangesServiceImpl.java
```diff
@@ -0,0 +1,114 @@
...
+public class UnpublishedChangesServiceImpl implements UnpublishedChangesService {
+
+    public boolean checkUnpublishedChanges(CoreSession session, DocumentModel document) {
+        
+        // Check that the document is a specific type using the helper method
+        if (!(checkType(document)))
+            return false;
+        
+        // Get the Workspaces document
+        String path = document.getPathAsString();
```
davidlacho: This is also a handy method that should be thrown into an abstract class that all services inherit from.

FirstVoicesPublisher/src/test/java/ca/firstvoices/operations/CheckUnpublishedChangesTest.java
```diff
@@ -0,0 +1,122 @@
...
+    dialectDoc.followTransition("Republish");
+    dialectDoc = session.saveDocument(dialectDoc);
+    ctx.setInput(dialectDoc);
+    assertFalse((Boolean) automationService.run(ctx, CheckUnpublishedChanges.ID));
+  }
+
+  /*
+        Helper method to create a dialect tree.
+     */
+  protected void createDialectTree() throws Exception {
```
davidlacho: Put this in an Abstract class for tests to avoid code reduplication

FirstVoicesPublisher/src/test/java/ca/firstvoices/services/UnpublishedChangesServiceImplTest.java
```diff
@@ -0,0 +1,120 @@
...
+        /*
+            Should now return false because the changes have been published.
+         */
+        dialectDoc.followTransition("Republish");
+        dialectDoc = session.saveDocument(dialectDoc);
+        assertFalse(unpublishedChangesServiceInstance.checkUnpublishedChanges(session, dialectDoc));
+
+    }
+
+    protected void createDialectTree() throws Exception {
```
davidlacho: Same as above!

FirstVoicesPublisher/src/test/java/ca/firstvoices/testUtil/MockStructureTestUtil.java
```diff
@@ -0,0 +1,168 @@
...
+import org.nuxeo.ecm.automation.OperationContext;
+import org.nuxeo.ecm.automation.OperationException;
+import org.nuxeo.ecm.core.api.*;
+
+import java.util.*;
+import java.util.stream.IntStream;
+
+import static org.junit.Assert.assertNotNull;
+
+public class MockStructureTestUtil {
```
davidlacho: I can't seem to find any usages of this. But this seems like a good class to turn into an abstract test class and then all tests can inherit from this. Handy methods can be thrown in here.

NolanVH:
@davidlacho The requested changes have been made.  Several of the tests have been changed to use abstract test utils classes. Thanks for the review!

PR #240, opened 2020-03-25
FW-1158: Code Cleanup

davidlacho:
First PR of code cleanup related to FW-1158.

- Potential Bug Fixes
- Reduce code duplication in tests
- General refactors


FirstVoicesSecurity/src/main/java/ca/firstvoices/securitypolicies/groups/LanguageAdministrators.java
```diff
@@ -32,7 +32,8 @@ public Access checkPermission(Document doc, ACP mergedAcp, NuxeoPrincipal princi
                 + principal.getName());
 
         // Skip administrators, system user and groups that aren't language administrators
-        if (additionalPrincipalsList.contains("administrators") || principal.equals(SecurityConstants.SYSTEM_USERNAME)
```
davidlacho: For context,  principal.equals(SecurityConstants.SYSTEM_USERNAME) will always evaluate false, as they are different types.

PR #249, opened 2020-04-02
FW-1134: Create distinction between lowercase and uppercase confusable characters

Dr-Slugma:
This branch involves creating a distinction between uppercase and lowercase confusable characters.

### DIFF
--- a/FirstVoicesData/src/main/java/ca/firstvoices/services/CleanupCharactersServiceImpl.java
+++ b/FirstVoicesData/src/main/java/ca/firstvoices/services/CleanupCharactersServiceImpl.java
@@ -70,6 +70,9 @@ public class CleanupCharactersServiceImpl extends AbstractService implements Cle
             if (uppercaseConfusableList != null) {
                 for (String confusableCharacter : uppercaseConfusableList) {
                     String characterTitle = (String) d.getPropertyValue( "fvcharacter:upper_case_character" );
+                    if (characterTitle.equals("")) {
+                        throw new FVCharacterInvalidException("Can't have uppercase confusable character if there is no uppercase character.", 400);
+                    }
                     if (confusables.put(confusableCharacter, characterTitle) != null) {
                         throw new FVCharacterInvalidException("Can't have confusable character " + confusableCharacter + " as it is mapped as a confusable character to another alphabet character.", 400);
                     }
--- a/FirstVoicesData/src/main/java/ca/firstvoices/services/CleanupCharactersServiceImpl.java
+++ b/FirstVoicesData/src/main/java/ca/firstvoices/services/CleanupCharactersServiceImpl.java
@@ -36,7 +36,9 @@ public class CleanupCharactersServiceImpl extends AbstractService implements Cle
 
         if (propertyValue != null) {
             Map<String, String> confusables = mapAndValidateConfusableCharacters(characters);
+            Map<String, String> uppercase_confusables = mapAndValidateConfusableCharacters(characters, "upper_case_confusable_characters");
             String updatedPropertyValue = replaceConfusables(confusables, "", propertyValue);
+            updatedPropertyValue = replaceConfusables(uppercase_confusables, "", updatedPropertyValue);
             if (!updatedPropertyValue.equals(propertyValue)) {
                  document.setPropertyValue("dc:title", updatedPropertyValue);
                 return document;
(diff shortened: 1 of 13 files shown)


FirstVoicesData/src/main/java/ca/firstvoices/services/CleanupCharactersServiceImpl.java
```diff
@@ -50,20 +50,23 @@ public DocumentModel cleanConfusables(DocumentModel document) {
...
-                    String characterTitle = (String) d.getPropertyValue("dc:title");
-                    if (confusables.put(confusableCharacter, characterTitle) != null) {
-                        throw new FVCharacterInvalidException("Can't have confusable character " + confusableCharacter + " as it is mapped as a confusable character to another alphabet character.", 400);
-                    }
-                    if (confusables.containsKey(characterTitle)) {
-                        throw new FVCharacterInvalidException("Can't have confusable character " + confusableCharacter + " as it is mapped as a confusable character to another alphabet character.", 400);
-                    }
-                    if (characterValues.contains(confusableCharacter)) {
-                        throw new FVCharacterInvalidException("Can't have confusable character " + confusableCharacter + " as it is found in the dialect's alphabet.", 400);
+            for (String confusable_set : field_names){
```
davidlacho, line 55: We should limit a nested loop here, as this increases complexity and this process is already very inefficient. Since we know the two field names we are trying to grab we should just access the properties directly on the document. 

This isn't super pretty but it could look something like this:

```java
    @Override
    public Map<String, String> mapAndValidateConfusableCharacters(List<DocumentModel> characters) throws FVCharacterInvalidException {
        Map<String, String> confusables = new HashMap<>();
        List<String> characterValues = characters.stream().filter(c-> !c.isTrashed()).map(c -> (String) c.getPropertyValue("dc:title")).collect(Collectors.toList());

        for (DocumentModel d : characters) {
            String[] lowerCaseConfusableList = (String[]) d.getPropertyValue("confusable_characters");
            String[] upperCaseConfusableList = (String[]) d.getPropertyValue("upper_case_confusable_characters");
            if (lowerCaseConfusableList != null) {
                for (String confusableCharacter : lowerCaseConfusableList) {
                    String characterTitle = (String) d.getPropertyValue("dc:title");
                    if (confusables.put(confusableCharacter, characterTitle) != null) {
                        throw new FVCharacterInvalidException("Can't have confusable character " + confusableCharacter + " as it is mapped as a confusable character to another alphabet character.", 400);
                    }
                    if (confusables.containsKey(characterTitle)) {
                        throw new FVCharacterInvalidException("Can't have confusable character " + confusableCharacter + " as it is mapped as a confusable character to another alphabet character.", 400);
                    }
                    if (characterValues.contains(confusableCharacter)) {
                        throw new FVCharacterInvalidException("Can't have confusable character " + confusableCharacter + " as it is found in the dialect's alphabet.", 400);
                    }
                }
            }
            if (upperCaseConfusableList != null) {
                for (String confusableCharacter : upperCaseConfusableList) {
                    String characterTitle = (String) d.getPropertyValue("upper_case_character");
                    if (characterTitle != null) {
                        if (confusables.put(confusableCharacter, characterTitle) != null) {
                            throw new FVCharacterInvalidException("Can't have confusable character " + confusableCharacter + " as it is mapped as a confusable character to another alphabet character.", 400);
                        }
                        if (confusables.containsKey(characterTitle)) {
                            throw new FVCharacterInvalidException("Can't have confusable character " + confusableCharacter + " as it is mapped as a confusable character to another alphabet character.", 400);
                        }
                        if (characterValues.contains(confusableCharacter)) {
                            throw new FVCharacterInvalidException("Can't have confusable character " + confusableCharacter + " as it is found in the dialect's alphabet.", 400);
                        }
                    }
                }
            }
        }
        return confusables;
    }

```

FirstVoicesData/src/main/resources/OSGI-INF/ca.firstvoices.studio.extensions.xml
```diff
@@ -5128,6 +5134,34 @@
                     </widget>
                 </subWidgets>
             </widget>
+            <widget name="upper_case_confusable_characters" type="list">
+                <labels>
+                    <label mode="any">Uppercase Confusable Characters</label>
+                </labels>
+                <translated>false</translated>
+                <helpLabels>
+                    <label mode="any">Any other uppercase characters that might be mistaken for this character. (Example: Ts' could
```
davidlacho, line 5143: Let's make this a little more obvious. something like `𝔍 could be mistaken for J`. 
This might help:
http://www.unicode.org/Public/security/12.0.0/confusables.txt

davidlacho:
Screen Shot 2020-04-06 at 4 30 31 PM (2)

There's an issue when choosing an uppercase confusable character when no uppercase character is defined.

PR #251, opened 2020-04-06
FW-1219: Add FVSnackbar for Physical Distancing update

davidlacho:
This ticket creates a Snackbar for displaying a COVID update. It displays on the front page.


davidlacho:
Waiting for @madebybridget to follow up with comms on messaging

PR #253, opened 2020-04-08
FW-917 - Searching in the top navbar then in the filter menu does not update search

coakenfold:
Removed the with-toggle HOC and related form data/methods
Reorganized component code:
- Moved propTypes annotation to the bottom just above Redux boilerplate
- Moved non-lifecycle methods after render()
- Removed commented out code


coakenfold:
Thanks for the review @davidlacho

PR #254, opened 2020-04-08
FW-1213 FW-1214 Add frontend and Cypress to docker-compose

NolanVH:
With these changes the docker configuration is now setup to create Cypress and apache frontend containers.

See the docker README changes for instructions on how to use the frontend and Cypress docker-compose configurations.


Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+RUN a2enmod headers && \
+a2enmod proxy && \
+a2enmod rewrite && \
+a2enmod proxy_http && \
+a2enmod ssl
+
+COPY --from=build /app/public /opt/fv/www/
+COPY docker/apache2/000-default.conf /etc/apache2/sites-enabled/000-default.conf
+
+ADD https://github.com/ufoscout/docker-compose-wait/releases/download/2.7.3/wait /wait
```
dy2288, line 49: What is the reason for this?

Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+        poppler-utils \
+        libwpd-tools \
+        exiftool \
+        ghostscript \
+        libreoffice \
+        apache2 \
+        libtcnative-1 \
+        ffmpeg \
+        gnupg2 \
+        ca-certificates \
```
dy2288, line 33: `ca-certificates` is one we could easily drop - it's for SSL on servers and this is just a local dev docker. Save us some time/package size.

Dockerfile
NolanVH: It is a docker-compose script that waits for ports before continuing to the next command on docker-compose up. In this case, on line 53 it runs the script and waits for the port specified in [frontend-docker-compose.yml](https://github.com/First-Peoples-Cultural-Council/fv-web-ui/blob/35020dc95c0b656bfe26f14ebef90084e0091a32/docker/frontend-docker-compose.yml#L9-L12) file. More details can be found [on the repo](https://github.com/ufoscout/docker-compose-wait).

docker/README.md
```diff
@@ -104,6 +104,22 @@ If you want to rerun the initial setup script for any reason, you can do so by r
...
+```
+
+To startup a copy of the frontend using apache, (not for frontend dev work), along with the backend run the following command in place of the command on step 3:
+```
+docker-compose -f docker-compose.yml -f frontend-docker-compose.yml up
+```
+To startup the frontend, backend, and run the complete Cypress test suite run the following command (requires :
+```
+docker-compose -f docker-compose.yml -f frontend-docker-compose.yml -f cypress-docker-compose.yml up --abort-on-container-exit
+```
```
davidlacho, line 121: @dy2288  @NolanVH  maybe add a comment that you can also use `-skip-tests` here

NolanVH: > @dy2288 @NolanVH maybe add a comment that you can also use `-skip-tests` here

I don't think that would make sense here because we would want this command to fail if the tests aren't passing. It also explains elsewhere in this readme where the -skip-tests flag can be used.

PR #259, opened 2020-04-16
Skip Java unit tests on ./setup_docker with -skip-tests flag

davidlacho:


docker/README.md
```diff
@@ -125,21 +131,26 @@ docker exec nuxeo-dev /bin/bash -c "nuxeoctl stop && nuxeoctl mp-install --accep
...
+
 #### Easy Way:
-* From the root of your module run the command ```./UpdateModule.sh```. If you are creating a new module you will need to copy the UpdateModule.sh script from another module into the root of your module
-(eg: copy ```/FirstVoicesData/UpdateModule.sh``` into ```/YourNewModule```).
 
-    Optionally you can add the flag ```-skip-tests``` to skip the compilation and running of unit tests during the module deploy (eg: ```./UpdateModule.sh -skip-tests```).
+- From the root of your module run the command `./UpdateModule.sh`. If you are creating a new module you will need to copy the UpdateModule.sh script from another module into the root of your module
+  (eg: copy `/FirstVoicesData/UpdateModule.sh` into `/YourNewModule`).
+
+      Optionally you can add the flag ```-skip-tests``` to skip the compilation and running of unit tests during the module deploy (eg: ```./UpdateModule.sh -skip-tests```).
```
NolanVH, line 140: It looks like the quote on this line is not working as intended.

PR #274, opened 2020-04-30
FW-1293 Fix autosuggest for categories only suggesting parent categories

NolanVH:
Autosuggest now correctly shows child and parent categories.

Credit to @coakenfold for the majority of the fix. Thanks Chad!


gmcauliffe:
Great work team. Tested locally.

PR #275, opened 2020-04-30
FW-1294: Unpublished changes based on diff

dy2288:
Notes:

- Document.CheckUnpublishedChanges: Rather than evaluating the version of documents, the service was updated to do a diff, excluding properties and schemas that are not likely to be updated by end-users. We will need to adjust this as we introduce new schemas that are not user facing.

- Document.GetUnpublishedChanges: This is a new end-point that will actually display the diff between the two documents.

This change introduces a new external library (https://javers.org/)

### DIFF
--- a/FirstVoicesPublisher/src/main/java/ca/firstvoices/services/UnpublishedChangesServiceImpl.java
+++ b/FirstVoicesPublisher/src/main/java/ca/firstvoices/services/UnpublishedChangesServiceImpl.java
@@ -1,62 +1,121 @@
 package ca.firstvoices.services;
 
 import ca.firstvoices.publisher.services.FirstVoicesPublisherService;
+import org.javers.core.Javers;
+import org.javers.core.JaversBuilder;
+import org.javers.core.diff.Diff;
 import org.nuxeo.ecm.core.api.CoreInstance;
-import org.nuxeo.ecm.core.api.DocumentModel;
-import org.nuxeo.ecm.core.api.DocumentRef;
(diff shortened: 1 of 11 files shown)


FirstVoicesPublisher/src/main/java/ca/firstvoices/services/UnpublishedChangesServiceImpl.java
```diff
@@ -1,62 +1,121 @@
...
+
+    // Helper method to get all relevant properties for the diff
+    private HashMap<String, Object> getDiffProperties(DocumentModel doc) {
+
+        // The following schemas shouldn't be considered in a diff (they are not user-controlled)
+        String[] excludedSchemas = {"uid", "collectionMember", "common", "fvancestry", "fvproxy", "fvlegacy", "relatedtext", "facetedTag"};
+        List<String> excludedSchemasList = Arrays.asList(excludedSchemas);
+
+        // The follow properties shouldn't be considered in a diff (they are not user-controlled)
+        String[] excludedProperties = {"dc:modified", "dc:created", "dc:contributors", "dc:lastContributor"};
```
davidlacho, line 89: @dy2288 Should we include the fields `fv-alphabet:update_confusables_required`, `fv-phrase:update_confusables_required`, and `fv-word:update_confusables_required`? These flags are used to perform scheduled operations.

dy2288: Yes, good point. Will add.

PR #278, opened 2020-05-01
FW-171 FW-1303 FW-1297 FW-1298 FW-1301 FW-1290 bug fixes

gmcauliffe:


davidlacho:
@gmcauliffe Other than the small UI bug we spoke about, PR LGTM 👍. Tested locally! Hoorah 🎉!

PR #280, opened 2020-05-01
FW-1295: Categories allowed for the creation of "Supercategories"

Dr-Slugma:
This only covers preventing a user from assigning a category that isn't the default 'Category' to be the parent of an existing category which has children.


gmcauliffe:
Tested locally and successfully catching attempts to assign parent categories as a child to another parent category.

PR #287, opened 2020-05-05
FW-1044: Bug Fix

davidlacho:


FirstVoicesEnrichers/src/main/java/ca/firstvoices/nuxeo/utils/EnricherUtils.java
```diff
@@ -330,11 +331,16 @@ public static String convertLetterToCustomOrder(CoreSession session, String dial
       // This is a good opportunity to queue a worker that will recalculate this in the future.
       // But for now let's try to just try to send over a "~ + letter":
       // TODO: Implement this in a way that ensures we always get the correct response.
-      DocumentModel dialectDoc = session.getDocument(new IdRef(dialect));
-      DocumentModel alphabet = session
-          .getDocument(new PathRef(dialectDoc.getPathAsString() + "/Alphabet"));
-      alphabet.setPropertyValue("fv-alphabet:update_confusables_required", true);
-      session.saveDocument(alphabet);
+
+      CoreInstance.doPrivileged(session, s -> {
```
dy2288, line 335: @davidlacho this will delay returning the result to the end user. Although it's a little more involved, I'd recommend firing an event (see https://github.com/nuxeo/nuxeo/blob/release-10.10/nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/main/java/org/nuxeo/ecm/automation/core/events/operations/FireEvent.java#L75) and then catching that event in a dedicated listener to update the flag. I don't believe you have to do that inside doPrivileged, but worth double checking. If you make the listener async, it will have little performance impact.

davidlacho: That's not a bad idea. I'm still on the fence about this being the best approach. I lean more towards computing native order as an intentional operation (not on read). But we need a value to return to the user.

FirstVoicesEnrichers/src/main/java/ca/firstvoices/nuxeo/utils/EnricherUtils.java
```diff
@@ -326,18 +324,8 @@ public static String convertLetterToCustomOrder(CoreSession session, String dial
...
-      // But for now let's try to just try to send over a "~ + letter":
-      // TODO: Implement this in a way that ensures we always get the correct response.
-      DocumentModel dialectDoc = session.getDocument(new IdRef(dialect));
-      DocumentModel alphabet = session
-          .getDocument(new PathRef(dialectDoc.getPathAsString() + "/Alphabet"));
-      alphabet.setPropertyValue("fv-alphabet:update_confusables_required", true);
-      session.saveDocument(alphabet);
-      return NativeOrderComputeServiceImpl.NO_ORDER_STARTING_CHARACTER + letter;
-    } else {
+      log.info("Dialect " + dialect + " requires custom order recompute.");
```
davidlacho, line 327: We actually shouldn't really ever be in the state after we run the compute custom order on all the dialects, and all future characters created will have a custom order field on them. So my solution instead of doing that write to the alphabet document is that in the rare case that the archive hasn't had the custom order recomputed we actually query the title "starts with," so the same behaviour as previous.

FirstVoicesEnrichers/src/main/java/ca/firstvoices/nuxeo/operations/GetDocumentsByCustomOrder.java
```diff
@@ -76,8 +77,12 @@ public DocumentModelList run() throws OperationException {
     if ((dialectId != null && !dialectId.isEmpty()) && (letter != null && !letter.isEmpty())) {
       customOrder = EnricherUtils.convertLetterToCustomOrder(session, dialectId, letter);
     }
-
-    if (customOrder.startsWith("%") || customOrder.startsWith("$") || customOrder.startsWith("&") ||
+    if (StringUtils.isEmpty(customOrder)) {
+      //      We shouldn't get to this state when the archive has the custom order computed on it.
```
davidlacho, line 81: We actually shouldn't really ever be in the state after we run the compute custom order on all the dialects, and all future characters created will have a custom order field on them. So my solution instead of doing that write to the alphabet document (see my previous commits on this branch) is that in the rare case that the archive hasn't had the custom order recomputed we actually query the title "starts with," so the same behaviour as previous.

FirstVoicesEnrichers/src/main/java/ca/firstvoices/nuxeo/utils/EnricherUtils.java
```diff
@@ -282,7 +280,7 @@ public static ArrayNode getRolesAssociatedWithDialect(DocumentModel doc, CoreSes
   public static String expandCategoriesToChildren(CoreSession session, String query) {
     if (query != null && !query.isEmpty()) {
       // Expand value of fv-word:categories so that it includes children
-      String REGEX = "(fv-word:categories|fvproxy:proxied_categories)\\/\\* IN \\(\"([a-zA-Z0-9\\-]*)\"\\)";
+      String REGEX = "(fv-word:categories|fvproxy:proxied_categories)/\\* IN \\(\"([a-zA-Z0-9\\-]*)\"\\)";
```
davidlacho, line 283: Escaping / is not necessary.

FirstVoicesEnrichers/src/main/java/ca/firstvoices/nuxeo/utils/EnricherUtils.java
```diff
@@ -76,7 +74,7 @@ private static String getThumbnailURL(DocumentModel doc, String xpath, String fi
   public static ObjectNode getBinaryPropertiesJsonObject(String binaryId, CoreSession session) {
 
     IdRef ref = new IdRef(binaryId);
-    DocumentModel binaryDoc = null;
+    DocumentModel binaryDoc;
```
davidlacho, line 77: Removed a bunch of redundant initializations to null.

FirstVoicesEnrichers/src/main/java/ca/firstvoices/nuxeo/operations/GetDocumentsByCustomOrder.java
```diff
@@ -53,7 +54,7 @@
   protected CoreSession session;
   @Context
   protected AutomationService automationService;
-  @Param(name = "query", required = true, description = "The query to " + "perform.")
+  @Param(name = "query", description = "The query to " + "perform.")
```
davidlacho, line 57: required = true is redundant.

PR #299, opened 2020-05-08
FW-1340 FW-1307 Maintenance package (with workers) 

dy2288:
This code has the framework for executing jobs on a type - it is currently implementing category operations to migrate categories (moving code from external utils to internal code base).


FirstVoicesData/src/test/java/ca/firstvoices/runner/FirstVoicesDataFeature.java
```diff
@@ -42,12 +42,13 @@
 @Deploy("org.nuxeo.ecm.platform.search.core")
 @Deploy("org.nuxeo.ecm.platform.webapp.types")
 
+// NOTE: All of FirstVoicesData may already be deployed? Not sure if this is needed.
```
davidlacho, line 45: I will make a ticket to look over unit tests deploys. These are getting out of hand.

davidlacho: https://firstvoices.atlassian.net/browse/FW-1369

FirstVoicesMaintenance/src/main/java/ca/firstvoices/dialect/RequiredJobsListener.java
```diff
@@ -0,0 +1,83 @@
...
+              if (dialects == null || dialects.size() == 0) {
+                return;
+              }
+
+              AutomationService automation = Framework.getService(AutomationService.class);
+              MaintenanceLogger maintenanceLogger = Framework.getService(MaintenanceLogger.class);
+
+              for (DocumentModel dialect : dialects) {
+
+                Set<String> requiredJobs = maintenanceLogger.getRequiredJobs(dialect);
```
davidlacho, line 57: I like that in the future we might be able to do code re-use by making this loop in an abstract class, as discussed.

FirstVoicesData/src/test/java/ca/firstvoices/runner/FirstVoicesDataFeature.java
dy2288: Thanks, also worth noting this package will need unit tests (and it probably won't be fun with all the async joy :D)

PR #309, opened 2020-05-13
FW-1235: Java Checkstyle Rules and Strict Enforcement

davidlacho:
Strict enforcement happens in root level pom.xml


davidlacho:
To merge after FW-1431. Expecting conflicts when that occurs. Transitioning PR to draft until that merge occurs.

PR #323, opened 2020-05-19
Cypress Best Practices

coakenfold:
- Moved logout command from afterEach to beforeEach. A Cypress test could be interrupted and tests that follow may fail since the afterEach logout would not have run - https://docs.cypress.io/guides/references/best-practices.html#Using-after-or-afterEach-hooks
- Removed out of date log messages in beforeEach
- Added some TODO notes to the afterEach wait command (related to video)


coakenfold:
Thanks @bronson-fpcc

PR #333, opened 2020-05-22
FW-1371: Parent categories able to be edited

davidlacho:


davidlacho:
Thanks for the review @gmcauliffe. It seems the only one that's not passing is the Jira Transition, and it wasn't pulling from the codestyle repo when GitHub had an outage. I have re-ran the checks and the build is passing. The Jira transitioning seems to be failing on all PRs because we don't have a column "DEV DONE" anymore, and I believe there's a ticket for that in the backlog.

PR #342, opened 2020-05-27
FW-1395 Add node status servlet

plasticviking:
# Changelog
 - Added `/nuxeo/nodestatus` servlet, returning a JSON object containing node cluster ID and instance name.
 - It is not possible to add detailed information to the `/nuxeo/runningstatus` endpoint, since it permits only pass/fail notifications.
 - Recommend this endpoint (and perhaps `/nuxeo/runningstatus`) be guarded by front proxy/balancer


davidlacho:
Changes LGTM 👍

PR #344, opened 2020-05-27
FW-1466 Checkboxes on Search Dialect need to be enabled & reset properly

coakenfold:
https://firstvoices.atlassian.net/browse/FW-1466


coakenfold:

davidlacho:
Will craft new pre-release with this bug fix. Thanks for getting this in speedily :)!

PR #347, opened 2020-05-28
FW-1404: cleanConfusablesOperations Test

bronson-c:
Added the cleanConfusablesChangesValues, a test to check that the value of each word/phrase that's run through the bulkservice is actually changed.


FirstVoicesData/src/test/java/ca/firstvoices/operations/CleanConfusablesOperationTest.java
```diff
@@ -72,9 +79,40 @@ public void callingOperationSetsPropertyOnAlphabet()
...
+    JSONBlob blob = (JSONBlob) automationService.run(ctx, CleanConfusablesOperation.ID);
+    Assert.assertNotNull(blob);
+    String id = blob.getString();
+
+    BulkService bulkService = Framework.getService(BulkService.class);
+    bulkService.await(id, Duration.ofMinutes(1));
+    String status = bulkService.getStatus(id).getState().toString();
+    Assert.assertEquals("COMPLETED", status);
+
+    DocumentModelList wordsAndPhrases = session
```
davidlacho, line 109: Although this is working for this test, If we change the way that we setup the mock unit test data, this can fail. I would recommend querying for the items the same way that the clean confusable operations test is.

FirstVoicesData/src/test/java/ca/firstvoices/operations/CleanConfusablesOperationTest.java
```diff
@@ -72,9 +79,40 @@ public void callingOperationSetsPropertyOnAlphabet()
...
+        .getChildren(new PathRef("/FV/Family/Language/Dialect/Dictionary"));
+
+    for (DocumentModel x : wordsAndPhrases) {
+      assertEquals(true, x.getPropertyValue("update_confusables_required"));
+    }
+
```
davidlacho, line 115: If you're up for strengthening this unit test, you can add a check that the operation ran as the expected user, as is the case with Nuxeo's unit tests on their bulk operations:

https://github.com/nuxeo/nuxeo/blob/master/modules/core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/bulk/TestSetPropertiesAction.java#L120-L122

FirstVoicesData/src/test/java/ca/firstvoices/operations/CleanConfusablesOperationTest.java
bronson-c: Changes the retrieval method to a query in the latest commit.

FirstVoicesData/src/test/java/ca/firstvoices/operations/CleanConfusablesOperationTest.java
bronson-c: Added a check to see if the service ran as the expected user in the latest commit.

PR #352, opened 2020-06-01
FW-1514, FW-1501, FW-1502

coakenfold:
https://firstvoices.atlassian.net/browse/FW-1514
Search: not saving settings on 1st search, subsequent searches work correctly

Cause/Background:
- /learn/words is considered different than /learn/words/10/1
- if you search from /learn/words, the results are shown on /learn/words/10/1
- during that transition the SearchDialect values are reset (cause of error)

Changes:
- updated urls to point to words & phrases with pagination included, eg: `/learn/words/10/1` instead of `/learn/words`
- added a check in `.../learn/phrases/list-view.js` to suppress duplicate requests. Fixes the 'old information is briefly displayed' issue

### DIFF
--- a/frontend/app/assets/javascripts/views/pages/explore/dialect/learn/base/index.js
+++ b/frontend/app/assets/javascripts/views/pages/explore/dialect/learn/base/index.js
@@ -605,7 +605,7 @@ export const sortHandler = async ({
   })
 }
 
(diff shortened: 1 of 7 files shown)


coakenfold:
Thanks for the review @elysse-fpcc !

PR #357, opened 2020-06-03
FW-1141: Sonarcloud Integration

elysse-fpcc:
I modified the POM file to use Jacoco's report-aggregate which is meant for multi-module projects.


davidlacho:
I timeboxed trying to get jacoco reports working. I'm closer, but nothing yet. Cleared to merge and we'll tackle that in a new ticket. Great work!

PR #361, opened 2020-06-04
FW-1487: Backend to add related assets to entry

davidlacho:


pom.xml
```diff
@@ -167,68 +167,6 @@
         <groupId>org.apache.maven.plugins</groupId>
         <version>3.1.1</version>
       </plugin>
-      <plugin>
```
davidlacho, line 170: These were added when trying to get unit tests to run to be scanned but are causing some failures in other packages.

davidlacho:
@dy2288 converting back to draft. I would like to add an enricher to facilitate FE development.

FirstVoicesEnrichers/src/main/java/ca/firstvoices/nuxeo/enrichers/WordEnricher.java
```diff
@@ -181,6 +181,31 @@ private ObjectNode constructWordJSON(DocumentModel doc) throws IOException {
...
+          try {
+            assetDoc = session.getDocument(ref);
+          } catch (DocumentNotFoundException | DocumentSecurityException de) {
+            continue;
+          }
+
+          ObjectNode assetObj = mapper.createObjectNode();
+          assetObj.put("uid", assetId);
+          assetObj.put("path", assetDoc.getPath().toString());
+          assetObj.put("dc:title", assetDoc.getTitle());
```
dy2288, line 203: Since these can be just about any document, should we include "type" here?

FirstVoicesOperations/src/main/java/ca/firstvoices/dialect/assets/services/RelationsServiceImpl.java
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * @author david
+ */
+public class RelationsServiceImpl implements RelationsService {
+
+  @Override
+  public DocumentModelList getRelations(CoreSession session, DocumentModel doc) {
+
+    String query = "SELECT * FROM Document WHERE fv:related_assets = '" + doc.getId() + "' AND "
```
dy2288, line 15: In order for the GetRelationsForAsset operation to work for both proxies and workspace documents, perhaps this query needs to be switched based on if `doc` is a proxy or a workspace (and then look at fvproxy field).

FirstVoicesEnrichers/src/main/java/ca/firstvoices/nuxeo/enrichers/WordEnricher.java
davidlacho: 100% Agree! I just did that in another commit :).

FirstVoicesOperations/src/test/java/ca/firstvoices/dialect/assets/services/RelationsServiceTest.java
```diff
@@ -0,0 +1,81 @@
...
+      session.saveDocument(word);
+    });
+
+    DocumentModelList assets = relationsService.getRelations(session, childCategory);
+
+    Assert.assertEquals(wordDocs.size(), assets.size());
+    wordDocs.forEach(word -> Assert.assertTrue(assets.contains(word)));
+  }
+
+  // This unit test is always returning one less document than expected and is failing.
```
davidlacho, line 51: The unit test for ca.firstvoices.dialect.assets.services.RelationsServiceTest#getProxiedRelationsForAsset is always returning one DocumentModel less than expected. So, considering a string of words (`String[] words = {"A", "B", "C", "D”, “E”, “F”};`), when calling the service with a proxy of a referenced word, it will only return 5/6 words in the DocumentModelList.
I have tried debugging and it seems that all the documents are receiving the proper reference in `fvproxy:proxied_related_assets`. I did this by querying `SELECT * FROM FVWord WHERE ecm:isVersion = 0 AND ecm:isProxy = 1` (to get all the proxies created), and then going through the documents and querying `document.getPropertyValue(fvproxy:proxied_related_assets)` to ensure that the reference to the proxied reference document is there. It seems to all be there (Gah!).

PR #363, opened 2020-06-08
FW-73 Addenda - Move user migration to Nuxeo

plasticviking:
Changelog
 - Added `fv.cognito.blacklistUsers` configuration option. Users in this list will always be authenticated locally and will never be migrated (use for Admin accounts). Comma-separated list.
 - Moved Cognito user migration into Nuxeo process.
 - Stripped unnecessary AWS SDK components from the build

Deployment notes:
 - Remove lambda migration trigger prior to deployment from within Cognito console


davidlacho:
Once https://github.com/First-Peoples-Cultural-Council/fv-web-ui/pull/361 is merged, the build should pass

PR #372, opened 2020-06-15
FW-73 Internal migration updates

plasticviking:
# Changelog
Updated previous PR to merge latest `upstream/master` and address edge cases highlighted in QA: Users will not be migrated if they do not have an email address, if their username is not an email address, or if their password does not meet the current password policy. In these cases, an `InvalidMigrationException` is thrown (and caught and logged at WARN level).


davidlacho:
lol @plasticviking "sonar boolean unboxing silliness" 😆

PR #378, opened 2020-06-18
Upgrade Material-UI to 4.10.2

coakenfold:
Changes based on:
https://material-ui.com/guides/migration-v3/


davidlacho:
Code looks good, @coakenfold! I'm stoked on this, as it will give us a lot more components to play with.

PR #384, opened 2020-06-21
FW-1343 Add categories endpoint

plasticviking:
# Changelog
- Add `FirstVoicesREST` module, which provides an additional `/nuxeo/api/v1/category` and `/nuxeo/api/v1/category/{uid}` endpoints.
- `/category` takes boolean param `publicOnly` (default true)
- Both take params `pageSize` and `currentPage` to paginate results
- Add `FirstVoicesRESTPageProviders` module to define and register the Page Providers. Nuxeo requires this to be in another module since it does not load extensions from OSGI fragments (and the REST endpoint must be an OSGI fragment to be patched into the Core `APIRoot`)


davidlacho:
Hey @plasticviking! Can we get unit tests to test the endpoints you provided using mock category data?

plasticviking:
> Looking good. Could you create a few unit tests for endpoint & response data & address sonar code smell. Thanks!

Tests created and sonar issues addressed. It took some time to find the correct incantations to appease Nuxeo.

PR #389, opened 2020-06-26
FW-1666 Words enabled for viewing in kids portal cannot be disabled

coakenfold:
The HOC was accidentally filtering out checkboxes===false


coakenfold:
Thanks for the review @bronson-c

PR #390, opened 2020-06-29
FW-1607 Tasks: Replace DictionaryList with Material UI Table

coakenfold:
Ticket originally was 'DictionaryList: Refactor for reuse, ease of use, & separation of concerns' and was meant to get DictionaryList ready to be used with Widgets. Opted instead to use the Material-UI Table library. Still applying principles of reuse & separation of concerns.

`/List` is essentially just a presentation layer but `/List/ListData` & `/List/ListContainer` have been included as examples. `/List/ListPresentation` will be used by other `List*` components (eg: `ListTasks`) and it is where Material-UI table is loaded and options initialized.

Added the new Material-UI table to the `/tasks` page.


coakenfold:

PR #405, opened 2020-07-06
FW-1648: Republish word on recompute if it's already published

davidlacho:


FirstVoicesPublisher/src/main/java/ca/firstvoices/nativeorder/services/NativeOrderComputeServiceImpl.java
```diff
@@ -160,15 +162,16 @@ protected void computeNativeOrderTranslation(DocumentModel[] chars, DocumentMode
       }
       CoreSession session = element.getCoreSession();
 
+      element.getCoreSession().saveDocument(element);
```
bronson-c, line 165: (nit) Would it be possible to just use the "session" variable created above instead of calling element.getCoreSession again?

PR #408, opened 2020-07-07
FW-1698 GitHub workflow addition for Jira

gmcauliffe:
Author Info comment added to Jira Issues upon PR - needs confirmation when next PR is made

### DIFF
--- a/.github/workflows/JiraTransition.yml
+++ b/.github/workflows/JiraTransition.yml
@@ -2,153 +2,181 @@ name: Jira Transition
 
 on:
   pull_request:
-    branches:
-      master
-    types: [opened, synchronize, reopened, edited, review_requested, review_request_removed]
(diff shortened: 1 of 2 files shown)


dy2288:
The integration with JIRA/PR info seems sufficient for me for now.

dy2288:
Notes from JIRA ticket:

Researched - can be added as comments similar to how we are adding Reviewers 
fv-web-ui/scripts/ci/ReviewerInfoComment.sh

Syntax for creating workflows
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions
Helpful link for knowing what can be accessed on different GitHub event objects
https://developer.github.com/v3/activity/event_types/
Created script including sh file. Couldn’t set the sh file to be executable. Needs debugging.

PR #410, opened 2020-07-08
FW-1512

coakenfold:
- Two components are in a single file but only one of them was Redux connected.
- Error triggered when filtering by shared resources where a nonexistent `this.props.intl` is called


coakenfold:
Thanks @gmcauliffe!

PR #415, opened 2020-07-13
FW-1708 Site crashes on word view w/small screens when on /sections

coakenfold:
- Triggered by accessing deeply nested object that is undefined when on `sections`
- Created a custom template to render State and enabled it over in `getColumns()`

### DIFF
--- a/frontend/app/assets/javascripts/views/components/Browsing/DictionaryListSmallScreen.js
+++ b/frontend/app/assets/javascripts/views/components/Browsing/DictionaryListSmallScreen.js
@@ -110,6 +110,7 @@ const DictionaryListSmallScreen = (props) => {
         }
       })
 
+      // First & Last Name
       templateData[firstLastName] =
         _firstName || _lastName ? (
           <div className={`${classNameDataItem} ${classNameDataItem}--firstNameLastName`}>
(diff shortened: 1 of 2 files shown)


frontend/app/assets/javascripts/views/components/Browsing/DictionaryListSmallScreen.js
```diff
@@ -296,9 +297,9 @@ export const dictionaryListSmallScreenColumnDataTemplateCustomAudio = ({ cellRen
   return cellRender ? <div className={`DictionaryListSmallScreen__audioGroup ${className}`}>{cellRender}</div> : null
 }
 
-// dictionaryListSmallScreenTemplateWords
+// dictionaryListSmallScreenColumnDataTemplateCustomState
 // --------------------------------------------------------------
-export const dictionaryListSmallScreenTemplateWords = ({ templateData }) => {
```
gmcauliffe, line 301: WordsListPresentation.js is using this export as well. It will need to be updated if this export is being renamed.

coakenfold: Hi @gmcauliffe thanks for the review. I could only find one instance of `export const dictionaryListSmallScreenTemplateWords` in the codebase

PR #423, opened 2020-07-23
FW 1613 - Two column dashboard

bronson-c:
Changes the current dashboard into one that has 2 columns, which merge if the screen is small.


coakenfold:
LGTM

PR #424, opened 2020-07-24
FW-1597 V2

coakenfold:
Adds the following features to Dashboard Widget > Tasks:
- pagination
- sort

Also upgraded font sizes to use a Typographic scale. Applies to Mat-UI's `<Typography />` component


coakenfold:
Thanks @gmcauliffe

PR #428, opened 2020-07-29
FW-1281 - Merge simplified REST API into Firstvoices

plasticviking:
# Changelog

This PR integrates the simplified API code from the `fv-api` repository completely into FirstVoices/Nuxeo, and removes all external dependencies (except Cognito for access control).

It uses page providers for customizable queries, and generates swagger documentation via  redoc for integration in the build artifact (viewable at http://host/nuxeo/redoc-static.html).

It additionally incorporates enhancements to FirstVoicesCoreTests to assist with testing  REST APIs and importing test data from YAML files.


FirstVoicesSimplifiedAPI/src/main/java/ca/firstvoices/simpleapi/model/AnnotationNuxeoMapper.java
```diff
@@ -0,0 +1,144 @@
...
+import java.util.Map;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+import javax.annotation.Nullable;
+import org.apache.commons.beanutils.PropertyUtils;
+import org.apache.commons.lang.reflect.FieldUtils;
+import org.nuxeo.ecm.core.api.DocumentModel;
+import org.nuxeo.ecm.core.api.PropertyException;
+
+public class AnnotationNuxeoMapper {
```
dy2288, line 22: @plasticviking my understanding is that this maps document models from Nuxeo to our API output format, but would be good to have a description in this class.

FirstVoices-marketplace/pom.xml
```diff
@@ -65,6 +70,10 @@
       <groupId>ca.firstvoices</groupId>
       <artifactId>FirstVoicesCoreTests</artifactId>
     </dependency>
+    <dependency>
```
dy2288, line 73: Remove duplicate declaration

FirstVoicesSimplifiedAPI/src/main/resources/OSGI-INF/extensions/ca.firstvoices.simpleapi.pageproviders.xml
```diff
@@ -0,0 +1,85 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<component name="ca.firstvoices.simpleapi.pageproviders">
```
dy2288, line 3: Is there a way to avoid duplication between this and the FirstVoicesRESTPageProviders package? Or leverage FirstVoicesREST so that calls are available to both the simple API and other parts of FirstVoices? Open to suggestions here.

plasticviking: There's not actually much duplication here, other than that both declare page providers. We could certainly standardize and keep them in one place.

The cleanest solution might be to have the FirstVoicesREST package calls the same service layer (combining the two modules and eliminating FirstVoicesPageProviders).

The two key differences between this and FirstVoicesREST are:
1. FirstVoicesREST goes through the normal Nuxeo security measures whereas the SimplifiedAPI is totally independent of the Nuxeo filter chain and relies on the JWT decoding baked into it.
2. FirstVoicesREST uses Nuxeo's object mapping (along with the enrichers, if desired), and the SimplifiedAPI uses a custom mapping to intermediate objects that are then mapped to JSON. It's this custom mapping that allows Swagger to create such detailed documentation about the response bodies (since it can reflectively inspect the Groovy classes and predict the structure of the responses).

PR #430, opened 2020-07-30
FW-1714 Add character validation and a back end field for ignored characters

bronson-c:
Adds the ability for ignored characters to be added to an alphabet in the back end and ensures the ignored characters are valid when added. Also adds a new method for character document validation.


FirstVoicesData/src/main/java/ca/firstvoices/listeners/FVDocumentListener.java
```diff
@@ -183,23 +183,46 @@ public void validateCharacter() {
...
+        //only test on update, not creation as characters will not exist during creation
+        if (event.getName().equals(DocumentEventTypes.BEFORE_DOC_UPDATE)) {
+          DocumentModelList characters = getCharacters(document);
+          DocumentModel alphabet = getAlphabet(document);
+          cleanupCharactersService.validateAlphabetIgnoredCharacters(characters, alphabet);
+
+        }
+
+      } catch (Exception exception) {
+        rollBackEvent(event);
```
davidlacho, line 221: Curious if there a particular reason why we're rolling back the event here (but not above)?

FirstVoicesData/src/main/java/ca/firstvoices/services/CleanupCharactersServiceImpl.java
```diff
@@ -37,7 +40,8 @@
 public class CleanupCharactersServiceImpl extends AbstractFirstVoicesDataService implements
     CleanupCharactersService {
 
-  private String[] types = {FV_PHRASE, FV_WORD};
+  private final String[] types = {FV_PHRASE, FV_WORD};
+  private static final String DOCUMENT_TITLE = "dc:title";
```
davidlacho, line 44: In the future, all dublincore (and other schema) property values should be kept in a constants file in FVData. Nothing to change here, just noting this for future developers.

FirstVoicesData/src/test/java/ca/firstvoices/listeners/FVDocumentListenerTest.java
```diff
@@ -72,4 +77,86 @@ public void testListener() {
...
+
+    List<String> testList = new ArrayList<>();
+    testList.add("a");
+    testList.add("b");
+
+    DocumentModel d = session.getChildren(alphabet.getRef()).get(0);
+    d.setPropertyValue("fvcharacter:upper_case_confusable_characters", (Serializable) testList);
+    session.saveDocument(d);
+  }
+}
```
davidlacho, line 162: I would like to see a passing test for ignored characters

FirstVoicesData/src/main/java/ca/firstvoices/services/CleanupCharactersService.java
```diff
@@ -30,4 +30,10 @@
   DocumentModel cleanConfusables(CoreSession session, DocumentModel document);
 
   Map<String, String> mapAndValidateConfusableCharacters(List<DocumentModel> characters);
+
+  boolean validateCharacters(List<DocumentModel> characters,
+      DocumentModel alphabet, DocumentModel updated);
+
+  boolean validateAlphabetIgnoredCharacters(List<DocumentModel> characters,
```
davidlacho, line 37: Please create unit tests for this interface method

FirstVoicesData/src/main/java/ca/firstvoices/services/CleanupCharactersServiceImpl.java
```diff
@@ -138,6 +143,141 @@ public DocumentModel cleanConfusables(CoreSession session, DocumentModel documen
     return confusables;
   }
 
+  @Override
+  public boolean validateCharacters(List<DocumentModel> filteredCharacters,
```
davidlacho, line 147: This validation method could be simplified as follows:

```java
  @Override
  public void validateAlphabetIgnoredCharacters(List<DocumentModel> characters,
      DocumentModel alphabet) {
    //This method only covers alphabets, characters are covered separately
    Set<String> collectedCharacters = createCharacterHashMap(characters);

    String[] ignoredCharsArr = (String[]) alphabet
            .getPropertyValue("fv-alphabet:ignored_characters");
    
    if (ignoredCharsArr != null) {
      for (String str : ignoredCharsArr) {
        if (!collectedCharacters.add(str)) {
          throw new FVCharacterInvalidException(
                  "The ignored characters list includes a duplicate character found in another character document",
                  400);
        }
      }
    }
  }
```

FirstVoicesData/src/main/java/ca/firstvoices/services/CleanupCharactersServiceImpl.java
```diff
@@ -65,27 +69,28 @@ public DocumentModel cleanConfusables(CoreSession session, DocumentModel documen
...
-        document.setPropertyValue("dc:title", updatedPropertyValue);
+        document.setPropertyValue(DOCUMENT_TITLE, updatedPropertyValue);
       }
     }
     document.setPropertyValue("fv:update_confusables_required", false);
     return session.saveDocument(document);
   }
 
   @Override
+  //Keeping method as it is used elsewhere in cleanConfusables
```
davidlacho, line 80: Since this is now a "helper" method, it should not be included in the interface and accessor level should be **private**. Likewise, you should refactor the unit tests for this method to include your new implementation.

FirstVoicesData/src/main/java/ca/firstvoices/services/CleanupCharactersServiceImpl.java
```diff
@@ -138,6 +143,141 @@ public DocumentModel cleanConfusables(CoreSession session, DocumentModel documen
...
+        if (!unique) {
+          lowerConfusable = false;
+          break;
+        } else {
+          lowerConfusable = true;
+        }
+      }
+    } else {
+      lowerConfusable = true;
+    }
```
davidlacho, line 177: Likewise, this can be cleaned up: 

```java
    String[] lowerConfusableStrArr = (String[]) updated
        .getPropertyValue("fvcharacter:confusable_characters");
    if (lowerConfusableStrArr != null) {
      for (String str : lowerConfusableStrArr) {
        if (!updatedDocumentCharacters.add(str)) {
//          Throw exception here.
      }
    }
```

FirstVoicesData/src/main/java/ca/firstvoices/services/CleanupCharactersServiceImpl.java
```diff
@@ -138,6 +143,141 @@ public DocumentModel cleanConfusables(CoreSession session, DocumentModel documen
...
+        if (!unique) {
+          upperConfusable = false;
+          break;
+        } else {
+          upperConfusable = true;
+        }
+      }
+    } else {
+      upperConfusable = true;
+    }
```
davidlacho, line 194: Refactor as recommended above.

FirstVoicesData/src/main/java/ca/firstvoices/services/CleanupCharactersServiceImpl.java
```diff
@@ -138,6 +143,141 @@ public DocumentModel cleanConfusables(CoreSession session, DocumentModel documen
...
+
+    if (!(lower && upper && lowerConfusable && upperConfusable)) {
+      throw new FVCharacterInvalidException(
+          "A character is duplicated somewhere in this document's uppercase, "
+              + "lowercase or confusable characters ",
+          400);
```
davidlacho, line 200: Throw the exception in the above validations.

FirstVoicesData/src/main/java/ca/firstvoices/services/CleanupCharactersServiceImpl.java
```diff
@@ -138,6 +143,141 @@ public DocumentModel cleanConfusables(CoreSession session, DocumentModel documen
...
+    Set<String> collectedCharacters = createCharacterHashMap(filteredCharacters);
+
+    if (alphabet
+        .getPropertyValue("fv-alphabet:ignored_characters") != null) {
+      collectedCharacters.addAll(Arrays.asList((String[]) alphabet
+          .getPropertyValue("fv-alphabet:ignored_characters")));
+    }
+
+    //Confirm that updated character set is unique from all other characters
+    boolean setsUnique = Collections.disjoint(updatedDocumentCharacters, collectedCharacters);
```
davidlacho, line 213: Similar cleanup:

```java
    if (Collections.disjoint(updatedDocumentCharacters, collectedCharacters)) {
      throw new FVCharacterInvalidException(
          "The updated character includes a duplicate character "
              + "found in another character document",
          400);
    }
```

FirstVoicesData/src/main/java/ca/firstvoices/services/CleanupCharactersServiceImpl.java
```diff
@@ -138,6 +143,141 @@ public DocumentModel cleanConfusables(CoreSession session, DocumentModel documen
...
+    if (!setsUnique) {
+      throw new FVCharacterInvalidException(
+          "The updated character includes a duplicate character "
+              + "found in another character document",
+          400);
+    }
+
+    return true;
```
davidlacho, line 221: method does not require a return value if you are throwing exceptions when validation fails.

FirstVoicesData/src/main/java/ca/firstvoices/services/CleanupCharactersServiceImpl.java
```diff
@@ -138,6 +143,141 @@ public DocumentModel cleanConfusables(CoreSession session, DocumentModel documen
...
+    }
+
+    if (!ignoredCharsUnique) {
+      throw new FVCharacterInvalidException(
+          "The ignored characters list includes a duplicate character "
+              + "found in another character document",
+          400);
+    }
+    return true;
+  }
```
davidlacho, line 255: Follow a similar refactor pattern as above. Validation does not require return value if throwing an exception in the validation method.

FirstVoicesData/src/main/java/ca/firstvoices/services/CleanupCharactersServiceImpl.java
bronson-c: Noted

FirstVoicesData/src/main/java/ca/firstvoices/listeners/FVDocumentListener.java
bronson-c: The only reson for the roll back event here is that the character updates and in a similar try-catch block. Unsure if it needs to change or not.

FirstVoicesData/src/main/java/ca/firstvoices/services/CleanupCharactersServiceImpl.java
bronson-c: Done

FirstVoicesData/src/main/java/ca/firstvoices/services/CleanupCharactersServiceImpl.java
bronson-c: Done

FirstVoicesData/src/test/java/ca/firstvoices/listeners/FVDocumentListenerTest.java
bronson-c: Done, only tested that ignored characters were set properly as the order recompute is tested in FirstVoicesPublisher

FirstVoicesData/src/main/java/ca/firstvoices/services/CleanupCharactersServiceImpl.java
bronson-c: Done

FirstVoicesData/src/main/java/ca/firstvoices/services/CleanupCharactersServiceImpl.java
bronson-c: Done. I did end up removing a mapAndValidateConfusableCharacters unit test as it was no longer a public method, but it is now only used in cleanConfusables so the unit test for cleanConfusables should suffice. All other tests utilizing mapAndValidateConfusableCharacters were converted to using validateCharacters.

FirstVoicesData/src/main/java/ca/firstvoices/listeners/FVDocumentListener.java
davidlacho: I don't think it's entirely necessary. On error you can just throw an exception and it will bubble up and be detected by the Nuxeo transaction layer which will roll back the current transaction.

PR #448, opened 2020-08-19
FW-1710 - New end points for tasks list view and details view 

dy2288:
There is some code that I kept "disabled" in FirstVoicesData as a reference for the future. Created tests for the API, but not marshalers (TODO).


FirstVoicesREST/src/main/java/org/nuxeo/ecm/restapi/server/jaxrs/firstvoices/SimpleTaskObject.java
```diff
@@ -0,0 +1,107 @@
...
+
+  @Override
+  protected PageProviderDefinition getPageProviderDefinition() {
+    return pageProviderService.getPageProviderDefinition(PAGE_PROVIDER_NAME);
+  }
+
+  @Override
+  protected Object[] getParams() {
+    // Get list of actors (user + group)
+    List<String> actors = TaskActorsHelper
```
dy2288, line 72: May need to expand this to a group (i.e. not just principal/username, but dene_lang_admins for example)

dy2288: Already does this in the getTaskActors.

PR #461, opened 2020-08-25
FW-1561 Add categories to mock data

bronson-c:
- Generate 9 categories in each test dialect, 3 parent 6 child
- Randomly assign words to a category upon creation


FirstVoicesPublisher/src/main/java/ca/firstvoices/templates/factories/FVDialectFactory.java
```diff
@@ -60,6 +60,7 @@ public void createContentStructure(DocumentModel eventDoc) {
         this.mockTemplate = new ArrayList<>(template);
```
dy2288, line 60: @bronson-c can you please add to the comment explaining why these are to be ignored for mock data structures?

bronson-c: Done

PR #465, opened 2020-09-02
FW-1825

elysse-fpcc:
- Removed "input" in CSS file as I realized it is not being used in this iteration of the component
- Renamed the classes in RequestChangesPresentation to follow the BEM convention
- FYI I accidentally included some of the older files for DetailSong but I removed them (this is why there are older commits)


coakenfold (approved):
Great! Thanks for updating the css/js. Approved!

PR #470, opened 2020-09-04
FW-1562 mock user service + operations

bronson-c:
Adds 4 new operations and a new service to mock data to add and remove users of each access level to already created dialects.


bronson-c:
Unfortunately won't have time to address sonar issues or review changes, however all the code is functional tested locally. Also code for removing users/groups is contained in this PR so please use it as reference for https://firstvoices.atlassian.net/browse/FW-1826 and add it to the dialect teardown

PR #472, opened 2020-09-09
FW-1816 Registration Widget 1st pass

coakenfold:
- Presentation layer coded
- Data Layer is using fake responses from server. Endpoint needs to be developed
- Updated Dashboard Presentation to iterate over a `widgets` prop


elysse-fpcc (approved):
Looks good @coakenfold

PR #473, opened 2020-09-09
FW-1624

coakenfold:
Before the new endpoint was available the Data layer grabbed the selected item's task details from a different endpoint.

Added `getSimpleTask(taskId)` to DataSource


gmcauliffe (approved):
LGTM 👍🏻

PR #478, opened 2020-09-10
FW-1624

coakenfold:
- Gets data for words or phrases
- Selected Item Panel: renders metadata & phrase books
- Fixed bug where selected item detail wasn't set on initial load
- Updated `RequestChanges*` to be able to set button text


coakenfold:
Note: The flagged code smell doesn't bother me. We'll probably be adding other item types in the near future.

gmcauliffe (approved):
LGTM 👍🏻

PR #481, opened 2020-09-11
FW-1624 Sorting on widget & detail view

coakenfold:
- Disabled sort on `Entry title` column since it's a deeply nested prop that we can't sort against (at the moment)
- Dashboard Detail > Table view updates url on sort
- `useDashboard` update to set the correct data prop name when sorting


elysse-fpcc (approved):
Changes look good to me

PR #483, opened 2020-09-15
FW-1838 - Create new endpoints for simple tasks

dy2288:
See postman/ticket for documentation.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,9 @@
 # Set the default behavior, in case people don't have core.autocrlf set.
 * text=auto
 
+# Ensure .sh files are always encoded as LF
+*.sh text
+
 # Denote all files that are truly binary and should not be modified.
 *.png binary
 *.jpg binary
--- a/FirstVoicesRESTPageProviders/UpdateModule.sh
+++ b/FirstVoicesRESTPageProviders/UpdateModule.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+# Get the path to this script
+DIRECTORY="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+# Pull the current module from the path
+MODULE=${DIRECTORY##*/}
+
+# Run the main UpdateModule script in the docker directory
+cd $DIRECTORY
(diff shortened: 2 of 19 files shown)


dy2288:
Note: Should have opened as a draft. Still need to create tests for end-points.

.gitattributes
```diff
@@ -1,6 +1,9 @@
 # Set the default behavior, in case people don't have core.autocrlf set.
 * text=auto
 
+# Ensure .sh files are always encoded as LF
```
dy2288, line 4: These files seem to be converted sometimes to CRLF, and not handled correctly despite `core.autocrlf` set appropriately. Would be best to ensure they are always converted to LF line endings.

FirstVoicesRESTPageProviders/UpdateModule.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
dy2288, line 1: RESTPageProviders was missing UpdateModule.sh

FirstVoicesSecurity/src/main/java/ca/firstvoices/securitypolicies/groups/LanguageRecorders.java
```diff
@@ -54,23 +54,21 @@
...
    */
-  private Boolean hasPermissionInACP(ACP mergedAcp, List<String> additionalPrincipalsList,
-      String permission) {
+  private boolean hasRecordPermissionInACP(ACP mergedAcp, List<String> additionalPrincipalsList) {
 
     for (ACL acl : mergedAcp.getACLs()) {
-      for (ACE ace : acl.getACEs()) {
-        if (ace.isGranted() && additionalPrincipalsList.contains(ace.getUsername()) && ace
-            .getPermission().equals(permission)) {
+      for (ACE ace : acl.getACEs()) { // get permission CAN be null
```
dy2288, line 69: Found a bug here as the Nuxeo task service will set some custom permissions on tasks, where the permission itself can be null. Rest of the changes here are SonarQube.

elysse-fpcc (approved):
The changes seem good to me. A couple things I was confused about but I think I answered all my questions after reading through everything. Approved 😸

PR #484, opened 2020-09-17
FW-1836 user feedback after action

coakenfold:
Originally after approve/reject user feedback was provided by a snackbar but I encountered problems with refreshing the list after processing an item. The refresh would re-render the snackbar and so it wouldn't display a message.

Had to figure out how to provide user feedback, indicate what had been processed, & prevent users from resubmitting processed items.

Ended up dropping the snackbar and displaying messages in the approve drawer

Also dropped an old covid snackbar in AppWrapper.js


dy2288 (approved):
Looks good to me! I would address the two sonarcloud issues and merge. It would be nice to work towards a more reusable and less repetitive pattern for this user feedback if this type of flow is going to be elsewhere (at the moment it isn't an issue, but something to think of)

coakenfold:
Thanks for the review!

The 2 sonarcloud issues relate to common Redux patterns. I addressed the switch statement issue since that component probably won't have more `action.type`s. Left the 'first arg has a default value' since that's what the Redux docs do.

PR #486, opened 2020-09-21
FW-1847 FW-1846

gmcauliffe:
- Added a Local date stringHelper
- fixed typo in the RequestReview snack-bar
- fixed logic in SongStoryPagesPresentation so that the label for literal translations only renders if there is a literal translation of the page


coakenfold (approved):
LGTM

PR #503, opened 2020-09-28
FW-1868 banner to advise user that there is an existing task on an entry

gmcauliffe:


frontend/app/assets/javascripts/views/hoc/view/with-actions.js
```diff
@@ -45,17 +47,20 @@ export default function withActions(ComposedFilter, publishWarningEnabled = fals
...
+                      computeEntity={this.props.computeItem}
+                      computePermissionEntity={this.props.permissionEntry}
+                      computeLogin={this.props.computeLogin}
+                      publishChangesAction={this._publishChangesAction}
+                      {...this.props}
+                    >
+                      &nbsp;
+                    </PageToolbar>
+                    {this._hasPendingReview(selectn('response', this.props.computeItem))}
+                  </div>
```
coakenfold, line 63: If this div is only for react (single ancestor constraint), maybe make it a fragment? eg: `<>` & `</>`. 
One less div in the client.

coakenfold (approved):
LGTM. 

One nitpick about considering react fragments instead of a div. Shouldn't block merging though.

Thanks for converting `<button>` to `<FVButton>` and removing `<br>`s

PR #504, opened 2020-09-29
FW-1875 - Ensure clean confusables does not run automatically after add confusables

dy2288:
A few other small fixes along the way (see commits below).


plasticviking (approved):
The code looks good to me

PR #509, opened 2020-10-01
FW-1856 task approval - republishing

gmcauliffe:
- Bug fix - Handle undefined item case in `hasPendingReview`
- Handle republishing when transitioning an entry to 'Public'
- Update frontend to use `Document.FollowLifecycleTransition`  instead of `FVPublish`


FirstVoicesPublisher/src/main/java/ca/firstvoices/publisher/listeners/ProxyPublisherListener.java
```diff
@@ -82,4 +78,22 @@ public void handleEvent(Event event) {
       service.setDialectProxies(doc);
     }
   }
+
+  /**
+   * @param transition transition event requested
+   * @param transitionFrom state transitioning from
+   * @return
+   */
+  private boolean publishing(String transition, String transitionFrom) {
```
dy2288, line 87: Use format isAction for these (isPublishing, isUnpublishing, isRepublishing) for better readability

PR #511, opened 2020-10-02
FW-1850 - remove tasks after changes to doc by non admins

dy2288:
Made some slight changes to FVCore, FVData.

Opted to put this logic in FVMaintenance so that we can use this for future post commit, async cleanup events that need to happen after a doc is committed. Alternatively this could like in an FVTasks package if we create one.


gmcauliffe (approved):
Checklist:
* can follow the flow of the code ✅ 
* Variables, methods are named properly and clearly ✅   NB: Could "PostCommitCleanup" be less abstractly named?)
* Methods are documented using comments (where not self-explanatory) ✅ 
* Any suggestions for refactoring? None  ✅ 
* Tests are created and cover at least 85% of the functionality ✅ 
* Code is unlikely to introduce new regressions ✅

PR #512, opened 2020-10-05
FW-1853 - Add view and edit buttons to detail view

elysse-fpcc:
- Added "edit" and "view item" buttons to the detail view of SongStory and WordPhrase components
- Displaying the buttons can be toggled on/off so they are only displayed on the dashboard version of each component


gmcauliffe (approved):
LGTM

PR #513, opened 2020-10-05
FW-1845 - Cleanup tasks and workflow code

dy2288:


gmcauliffe (approved):
Checklist:

can follow the flow of the code ✅
Variables, methods are named properly and clearly ✅ 
Methods are documented using comments (where not self-explanatory) ✅
Any suggestions for refactoring? None ✅
Tests are created and cover at least 85% of the functionality ✅
Code is unlikely to introduce new regressions ✅

NB: We will need to make a ticket to remove all uses of `workflow` in the Frontend - 20 odd files and `actions`

PR #515, opened 2020-10-05
FW-1906 - include current state in core entity output; slight refactors

dy2288:


gmcauliffe (approved):
* Can follow the flow of the code ✅
* Variables, methods are named properly and clearly ✅ 
* Methods are documented using comments (where not self-explanatory) ✅
* Any suggestions for refactoring? None ✅
* Tests are created and cover at least 85% of the functionality ✅
* Code is unlikely to introduce new regressions ✅

PR #516, opened 2020-10-06
FW-1894

coakenfold:
Fixes display issues on Dashboard Detail when a user account is missing first & last name. Uses email as a fallback.

Before:
BEFORE

After:
AFTER


gmcauliffe (approved):
LGTM

PR #517, opened 2020-10-06
FW-1898 requested visibility optional

gmcauliffe:
Change frontend to only send a `requestedVisibility` when there is a change requested


frontend/app/assets/javascripts/components/RequestReview/RequestReviewData.js
```diff
@@ -80,8 +80,10 @@ function RequestReviewData({ children, docId, docState, docType }) {
 
   const handleDialogOk = () => {
     setIsDialogOpen(false)
+    // If no change to visibility requested, set requestedVisibility to null
+    const requestedVisibility = requestVisibilityType === docVisibility ? null : requestVisibilityType
     // Send request to the server to set related task for the document
-    postRequestReview(docId, requestVisibilityType, comment)
+    postRequestReview(docId, requestedVisibility, comment)
```
coakenfold, line 86: Using a single arg would eliminate things like:
```
const docId = '123-123'
const requestedVisibility = null
const comment = 'some comment'
postRequestReview(docId, requestedVisibility, comment)
```

You could do:
```
postRequestReview({docId, comment})
```

frontend/app/assets/javascripts/providers/redux/reducers/tasks/actions.js
```diff
@@ -110,6 +110,21 @@ export const postRequestReview = (docId, requestedVisibility, comment) => {
...
+        .then((response) => {
+          dispatch({ type: FV_REQUEST_REVIEW_POST_SUCCESS, document: response })
+          return response
+        })
+        .catch((error) => {
+          dispatch({ type: FV_REQUEST_REVIEW_POST_ERROR, error: error })
+          return error
+        })
+    }
+
```
coakenfold, line 127: Wonder if we could drop this duplicate block `if(!requestedVisibility) { /* ... */ }`?

Maybe something like:
```
const xhrObj = {docId, comment}
if (requestedVisibility) { xhrObj.requestedVisibility = requestedVisibility }
return DirectoryOperations.postToAPI(`${URLHelpers.getBaseURL()}api/v1/simpleTask/requestReview`, xhrObj)
```

coakenfold (changes requested):
Think the duplication in frontend/app/assets/javascripts/providers/redux/reducers/tasks/actions.js should be addressed

frontend/app/assets/javascripts/components/RequestReview/RequestReviewData.js
coakenfold: ^ just a FYI not a change request

coakenfold (approved):
Lovely! Thanks for the revision. LGTM

PR #520, opened 2020-10-06
FW-1896 Bug fix - not saving emptied form fields

gmcauliffe:
Edit form was not saving empty fields, so for example, if a user added pronunciation to a word, they could subsequently update the pronunciation but they couldn't delete it entirely.

NB: `getValue()` on tcomb-form is returning `null` values for string fields that are submitted as empty

Removed `null` filtering logic for non-list form values in _onRequestSaveForm function


coakenfold (approved):
LGTM

I believe the `// Note` on line 70 refers to unchecking checkboxes (eg: removing a word from the children's section). We should remember to test it.

PR #528, opened 2020-10-07
FW-1915

coakenfold:
After requesting changes the feedback text used to say:"We shared your feedback with the member"

It has been changed to:`Changes have been requested from ${fullName || email || 'the user'}`


gmcauliffe (approved):
LGTM :)

PR #537, opened 2020-10-15
FW-1921

coakenfold:
JIRA: https://firstvoices.atlassian.net/browse/FW-1921

- Links Widget outputs sets of links for each language a user administers. Previous version would only output the first language.
- Deleted WidgetRegistrations since wasn't being used.

Before:
Before

After:
After


gmcauliffe (approved):
LGTM 👍🏻

PR #539, opened 2020-10-21
FW-1914 - cleanup/refactoring of publishing

dy2288:


FirstVoicesMaintenance/src/main/java/ca/firstvoices/maintenance/listeners/PostCommitCleanupListener.java
```diff
@@ -71,7 +81,7 @@ public void handleEvent(Event event) {
    *
    * @return list of all events to handle; modify to handle more events.
```
plasticviking, line 82: returns a set, not a list.

plasticviking (approved):
Looks good to me

PR #563, opened 2020-10-29
FW-1924 Enable caching of the portals list API response

plasticviking:
# Changelog
There are many potential performance improvements, but this one I think is a good step in the right direction. It will cause the client to cache the `get_portals` response indefinitely, revalidating with the server on each request. This also gives an easy path to implement more sophisticated server-side invalidation in the future.


FirstVoicesRESTPageProviders/src/main/resources/OSGI-INF/extensions/ca.firstvoices.pageproviders.portals.xml
```diff
@@ -0,0 +1,27 @@
...
+
+  <require>org.nuxeo.runtime.started</require>
+
+  <extension target="org.nuxeo.ecm.platform.query.api.PageProviderService"
+             point="providers">
+
+
+    <genericPageProvider name="PORTALS_LIST_SECTIONS_PP" class="org.nuxeo.elasticsearch.provider.ElasticSearchNxqlPageProvider">
+      <property name="checkQueryCache">true</property>
+      <pattern quoteParameters="false">SELECT * FROM FVPortal WHERE ecm:path STARTSWITH '/FV/Sections' AND ecm:isTrashed
```
dy2288, line 13: I think `Sections` has to be lower case -> `sections`.

FirstVoicesREST/src/main/java/org/nuxeo/ecm/restapi/server/jaxrs/firstvoices/PortalsObject.java
```diff
@@ -0,0 +1,74 @@
...
+import javax.ws.rs.core.CacheControl;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import org.apache.http.HttpHeaders;
+import org.nuxeo.ecm.core.api.DocumentModel;
+import org.nuxeo.ecm.webengine.model.WebObject;
+import org.nuxeo.ecm.webengine.model.impl.DefaultObject;
+
+@WebObject(type = "portal")
```
dy2288, line 20: We've renamed `archive` / `dialect` / `portal` to `sites` - so might make sense to rename the wording here and in the methods (although the type will still be FVPortal)

dy2288 (changes requested):
Good to go after changes - excited to see this in action.

(Side note: It looks like we still need to work on invalidating the cache based on related items btw (e.g. related_audio)?)

PR #566, opened 2020-10-30
Enabling sourcemaps generation, dropping unused deps

coakenfold:
frontend/build/webpack.common.js
- fix invalid `devtool` value

frontend/build/webpack.development.js
- eslint fixes

frontend/build/webpack.production.js
- fix invalid `devtool` value

frontend/build/webpack.production.sourcemap.js
- Settings to generate sourcemaps.

frontend/package.json
- Added `npm run production:sourcemap` Takes a bit of time to generate sourcemaps so it's not part of `npm run production`
- Dropped dependencies that weren't being used in the codebase
- Had to bump up `eslint` related plugins due to a dependency error I was getting during the precommit hook.


gmcauliffe (approved):
Great to see we could drop all those dependencies. Thanks, Chad!

PR #571, opened 2020-11-03
FW-1979 html tags rendering in songs/stories

gmcauliffe:
Prevent user from seeing html tags in page and form:
- DOMPurify added to SongStoryPagesPresentation translations
- Wysiwyg added to form


elysse-fpcc (approved):
Looks good 👍

PR #572, opened 2020-11-04
FW-2002 - create new operation context within loop

dy2288:
Should solve issues with log files reporting "cannot find operation with input type void" when multiple required jobs are present.


plasticviking (approved):
Looks good

PR #573, opened 2020-11-04
FW-1949

coakenfold:
- Created a file that loads & sets options for DomPurify.
- Updated components that used to directly import DOMPurify.
- SongsStories/entry/view: Dropped the <Cover /> component. Don't think it was being used.


gmcauliffe (approved):
LGTM

coakenfold:

PR #574, opened 2020-11-04
FW-2001 - Publishing fixes

dy2288:
Issues resolved:

- Listener was triggering for proxies, as well as workspaces (wrong condition in ProxyPublisherListener), although it most likely did not affect functionality, it would have produced some error in log files
- Dialects can reference shared links (from SharedData) in Keyboards/language resources - this was the main reason for publishing not working in Production, which we didn't catch in Dev/Preprod
- Optimization (and possible solution for things stuck in Republish): the republish end-point will always explicitly set the document to publish bypassing the lifecycle service, and thus not triggering an additional event.

### DIFF
--- a/FirstVoicesMaintenance/src/main/java/ca/firstvoices/maintenance/listeners/ExecuteRequiredJobsListener.java
+++ b/FirstVoicesMaintenance/src/main/java/ca/firstvoices/maintenance/listeners/ExecuteRequiredJobsListener.java
@@ -116,7 +116,7 @@ public class ExecuteRequiredJobsListener implements EventListener {
                     event.markBubbleException();
                     log.severe(
                         () -> "Failed when trying to execute required jobs with message: " + e
-                            .getMessage());
+                            .getMessage() + " on dialect " + dialect.getId());
                   }
                 }
               }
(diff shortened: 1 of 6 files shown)


FirstVoicesMaintenance/src/main/java/ca/firstvoices/maintenance/listeners/ExecuteRequiredJobsListener.java
```diff
@@ -117,7 +117,7 @@ public void handleEvent(Event event) {
                     event.markBubbleException();
                     log.severe(
                         () -> "Failed when trying to execute required jobs with message: " + e
-                            .getMessage());
+                            .getMessage() + " on dialect " + dialect.getId());
```
gmcauliffe, line 120: Great idea. This will be helpful 👍🏻

gmcauliffe (approved):
* Can follow the flow of the code ✅
* Variables, methods are named properly and clearly ✅ 
* Methods are documented using comments (where not self-explanatory) ✅
* Any suggestions for refactoring? None ✅
* Tests are created and cover at least 85% of the functionality ✅
* Code is unlikely to introduce new regressions ✅

PR #580, opened 2020-11-05
FW-1831

coakenfold:
- `components/Alphabet/view` extracted to `components/AlphabetDetail` and refactored to the layered style 
- Added an enricher to return videos in the character section


dy2288 (approved):
Looks good Chad! Haven't tested, but code seems sound (both FE/BE)

PR #581, opened 2020-11-09
FW-1922: Speed improvements, bundle splitting for caching

dy2288:
See notes in ticket


coakenfold (approved):
Cool! Wondering what was happening with the `<Navigation />` component.

Approved (optional nitpick would be to minify the new inline css/js in `index.html`).

dy2288:
Thanks Chad.

> Cool! Wondering what was happening with the `<Navigation />` component.

It's not needed with the Navigation extracted to HTML, and login component the only react component in the menu.

> Approved (optional nitpick would be to minify the new inline css/js in `index.html`).

I tried looking at minifying the css/js via web pack extensions, but that required some changes to dependencies and didn't want to get into a rabbit hole. The different in speed seems negligible.

PR #622, opened 2020-12-02
FW-2015, FW-2070, FW-2063 - various explore language / state bugs

dy2288:
FW-2015 - Remove default="false" from fvm:child_focused. Does not seem to have any side effects. Since we use this schema on proxies, and the system seems to attempt to set `false` on update, publishing results in a `cannot set property on proxy` error. See also https://jira.nuxeo.com/browse/NXP-19466

FW-2070 - Ensure Enable, Disable, and RevertToNew apply to children when changing a dialect (caused issues with members not seeing their dialect)

FW-2063 - Workspaces query does not need language family title (explore language). Caused issues by non-super admins trying to access the language object which they do not have permission for.

#### Definition of Done


elysse-fpcc (approved):
Looks good, thanks!

PR #623, opened 2020-12-03
FW-1993

dy2288:
Hover issue - see https://stackoverflow.com/questions/59470947/vue-cli-build-production-change-opacity-in-css

Added "groups" to `sections` end-point so that it can still be cached server side, but the groups of the logged in user can be used to determine whether they have READ access to it or not.

#### Definition of Done


elysse-fpcc (approved):
Frontend changes look good :)

PR #627, opened 2020-12-04
FW-2023 Mock Data Improvements

elysse-fpcc:
3 main improvements:
- Made roughly 1/2 of words and phrases available in the kids portal and 1/3 of words available in games
- Added "definition" field to words and phrases and added tests to check for this field
- For generate mock users, I added the user's role in the first name field so that it is easy for us to find

#### Definition of Done
- [x] Variables, methods are named properly and clearly
- [x] Methods are documented using comments (where not self-explanatory)
- [x] Any suggestions for refactoring have been implemented
- [x] Tests are created and cover at least 85% of the functionality
- [x] Code is unlikely to introduce new regressions


gmcauliffe (approved):
LGTM

modules/common/firstvoices-core-tests/src/main/java/ca/firstvoices/tests/mocks/services/MockDialectServiceImpl.java
```diff
@@ -324,12 +324,29 @@ public DocumentModelList generateFVWords(CoreSession session, String path,
...
+      definitionEntry.put(LANGUAGE, ENGLISH);
+      definition.add(definitionEntry);
+      wordDoc.setPropertyValue("fv:definitions", definition);
       wordDoc.setPropertyValue("fv-word:part_of_speech",
           samplePartsOfSpeech[ThreadLocalRandom.current().nextInt(0, samplePartsOfSpeech.length)]);
       wordDoc.setPropertyValue("fv-word:pronunciation", wordDoc.getName() + " pronunciation");
 
+      //Makes the word available in kids portal with 1/2 chance
+      if (ThreadLocalRandom.current().nextInt(0, 2) == 0) {
+        wordDoc.setPropertyValue("fv:available_in_childrens_archive", "true");
```
gmcauliffe, line 342: Should this be setting the value to `true` rather than `"true"`?

elysse-fpcc:
Thanks for the review @gmcauliffe, looked into it a bit and in this case I can use either "true" or true. I opted to go for the boolean because I think it makes more sense here.

PR #633, opened 2020-12-11
FW-2085 Add bulk update operation

plasticviking:
Adds a new operation, `Document.BulkPropertyUpdate` callable from code or via the command API endpoint.

Accepts a set of DocumentRefs, a mode enum value defining whether to update/replace values or add values to an existing array property, a property name, and the value to update or add.

#### Definition of Done
- [x] Reviewer can follow the flow of the code
- [x] Variables, methods are named properly and clearly
- [x] Methods are documented using comments (where not self-explanatory)
- [x] Any suggestions for refactoring have been implemented
- [x] Tests are created and cover at least 85% of the functionality
- [x] Code is unlikely to introduce new regressions


modules/common/firstvoices-operations/src/main/java/ca/firstvoices/operations/bulkupdate/services/BulkUpdateServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class BulkUpdateServiceImpl implements BulkUpdateService {
+
+  public void bulkUpdate(
+      CoreSession session, DocumentRefList refs, BulkUpdateMode mode, String field,
+      Serializable value) {
+
+    for (DocumentRef ref : refs) {
+      DocumentModel doc = session.getDocument(ref);
+      if (doc == null) {
```
dy2288, line 21: Checking for existence here wouldn't hurt -- session.exists -- just in case one user deleted a document and another is trying a bulk operation on that document

dy2288 (changes requested):
See small note about "exists", otherwise - looks good to merge!
We should consider https://doc.nuxeo.com/nxdoc/bulk-action-framework/ for operations that select a lot of documents - maybe via a query or page provider?

PR #642, opened 2021-01-08
FW-2140 Site Endpoint

plasticviking:
Created `/api/v1/site/sections/{sitename}` endpoint.

A single site entity will be returned, or a 404 if there is no match. {sitename} will be matched in this order, with the first match winning. Matches must be exact.

- fvdialect:short_url
- ecm:name
- dc:title
- dialect uuid

--

Also addresses FW-2145 caching issue (tested locally).

#### Definition of Done


dy2288 (approved):
Looks good! I like the fallback mechanism with page providers!

PR #643, opened 2021-01-11
FW-2056 v2

coakenfold:
Upgrade to Webpack 5 (making Federated Modules available) plus code splitting is now enabled.

Highlights:
- Webpack config updates: config options updated to new spec, removed expose-loader (related to games lib)
- Package updates: updated script commands, dependency versions, 
- Updated babel.config: removed old plugins
- Changes so that routes/pagesIndex uses `React.lazy`
- Moved game library imports to games files. Allowed us to eliminate the extra `games_lib` entry point/bundle
- CSS imports no longer need: `!style-loader!css-loader!`
- Added the Suspender component that will reduce duplicate `<Suspense>` tags throughout the codebase
- Bootstrap.css added to index.html with a non-blocking loading system
- Replaced deprecated `request` http request library with `ky`

#### Definition of Done


frontend/package.json
```diff
@@ -72,118 +64,115 @@
...
+    "compression": "^1.7.4",
+    "copy-webpack-plugin": "^7.0.0",
+    "copyfiles": "^2.4.1",
+    "css-loader": "^5.0.1",
     "cypress": "^6.2.0",
     "cypress-file-upload": "^4.1.1",
     "cypress-pipe": "^2.0.0",
-    "del": "~1.2.1",
-    "eslint": "^7.10.0",
+    "dom-testing-library": "^5.0.0",
```
gmcauliffe, line 140: Testing-library has deprecated the old syntax for declaring their dependencies. See line 124, 125 vs 140 and 164. 
The new syntax is as follows:
"@testing-library/dom": "^7.29.1",
"@testing-library/react": "^11.2.2",

frontend/babel.config.js
```diff
@@ -1,35 +1,10 @@
...
-    'syntax-dynamic-import',
-    ['@babel/plugin-proposal-decorators', { legacy: true }],
-    'dynamic-import-node',
-    '@babel/plugin-transform-runtime',
-  ]
-
   return {
-    presets,
-    plugins,
+    // TODO: FW-2056 - Are we using the env.test config?
```
gmcauliffe, line 4: Not to my knowledge. @dy2288?

gmcauliffe (approved):
See package.json comment, but otherwise LGTM 👍🏻

frontend/babel.config.js
dy2288: Not sure - but this is the commit where it was added: https://github.com/First-Peoples-Cultural-Council/fv-web-ui/commit/23d8812355fa60071d9dabce0c4f00edc6b81d80#diff-fde68b13f2128c349915e8966c52c978913559b3777cb5dc1bf345b735d2f1d0

If removing it doesn't impact Cypress/Jest, we may be good to do it.

frontend/package.json
coakenfold: Thanks for catching this. Updated the package.

frontend/babel.config.js
coakenfold: I just resolved an issue with Jest tests not working. Was in a situation where adding presets/plugins would get the tests running but the site would break. 

`env.test` is used when Jest is running (and probably Cypress) so I added presets/plugins Jest needs in order to load up our components (without also breaking the site).

PR #649, opened 2021-01-15
Add categories and phrasebooks endpoints

plasticviking:
#### Definition of Done


dy2288 (changes requested):
Looks good @plasticviking but a few changes/additions:
* You can just have one version of categories (the new one), so please make v2 the only version
* I noticed that the entry count for categories doesn't include subcategories, so you need an IN query that includes both the parent and children (`fv-word:categories/* IN ('UID', 'UID')`). Phrase books are flat (no children), but categories can have children one level deep. This is done btw in EnricherUtils -> expandCategoriesToChildren as well.
* Also consider that for this to work with `sections` the FIND_WORDS_WITH_CATEGORY_PP and FIND_PHRASES_WITH_PHRASEBOOK_PP should be looking at `fvproxy:proxied_categories` when the dialect is a proxy/in sections
* One good addition to have is a `parentId` for child categories, which can be null (or missing) if the category is a parent itself.

Thanks!

plasticviking:
> Looks good @plasticviking but a few changes/additions:
> 
> * You can just have one version of categories (the new one), so please make v2 the only version
> * I noticed that the entry count for categories doesn't include subcategories, so you need an IN query that includes both the parent and children (`fv-word:categories/* IN ('UID', 'UID')`). Phrase books are flat (no children), but categories can have children one level deep. This is done btw in EnricherUtils -> expandCategoriesToChildren as well.
> * Also consider that for this to work with `sections` the FIND_WORDS_WITH_CATEGORY_PP and FIND_PHRASES_WITH_PHRASEBOOK_PP should be looking at `fvproxy:proxied_categories` when the dialect is a proxy/in sections
> * One good addition to have is a `parentId` for child categories, which can be null (or missing) if the category is a parent itself.
> 
> Thanks!

I've updated the code to match your suggestions. Should be ready for a final review and merge now.

dy2288 (approved):

PR #651, opened 2021-01-20
FW-2206

coakenfold:
Enables Federated Module sharing between two sites (V1 & V2). An example header is exported from V2 and loaded in V1

V2 Features:
- can control the polyfills that are included by the browsers we support (set in webpack configs)
- normalize/sanitize css files included via postcss. Can control the declarations that are included via package.json > browserslist setting

#### Definition of Done


gmcauliffe (approved):
LGTM

PR #658, opened 2021-01-27
FW-2112, FW-2055

coakenfold:
FW-2112:
- Adds About component
- Adds Content component (good for text)
- Adds Hero component
- Adds Suspender component (from V1)
- Adds V2 Jest testing. Opted to co-locate Jest tests since they are unit tests it makes more sense to group them with the component and you can see if something hasn't been tested.
- Adds useSearchParams hook to return search params as an object with parsed values
- Adds support for NUXEO_URL env var

FW-2055:
- Adds `services/api` for network requests
- Adds react-query

#### Definition of Done


gmcauliffe (approved):

dy2288:
@coakenfold is `NUXEO_URL` variable required now for v2, or is that just for testing? If it's required we'll need to update our dev/preprod/prod build scripts or find an alternative to get it.

coakenfold:
@dy2288 I'm not committed to the current implementation. I was referencing a system in place over in V1. It's used here: https://github.com/First-Peoples-Cultural-Council/fv-web-ui/blob/1a9738928257aed73e713c800daaa60b6592eb7a/frontend/app_v2/src/services/api/index.js#L11

The sticking point is the port number is different when trying to call nuxeo from V2

PR #659, opened 2021-01-28
FW-2212 - changes to serve dual app with apache2 in docker

dy2288:
A few other changes:
* Setup docker with `--no-cache` flag so that latest distributions are actually downloaded
* Assembly.xml - fix issue with build locally. Only include AWS jars if profile is "full-marketplace"
* Fix issue with login locally in cognito-config.xml.nxftl (do not use default if Cognito is set to false)
* FE: No need to include V2 header in V1 - moved to debug page
* Generated `.js` files with commit hash for decent cache busting (across both apps)
* Added git revision info to HTML
* Renamed `npm run build:evergreen` to  `npm run build:production` for consistency (we should consider removing the legacy tasks in V2)
* Added support for env variables in webpack build
* Dealt with postcss bug: https://github.com/tailwindlabs/tailwindcss/discussions/2462

#### Definition of Done


coakenfold (approved):
LGTM

PR #673, opened 2021-02-03
Add typeadapters to accept json inputs for bulk update operation

plasticviking:
#### Definition of Done


plasticviking:
This fixes the issues encountered when trying to call the operation via a web endpoint (which were related to Nuxeo not knowing how to (de)serialize the relevant types).

`mode` input must be provided as either "UPDATE_PROPERTY" or "ADD_TO_ARRAY_PROPERTY" when calling via HTTP.

dy2288 (changes requested):
Code looks sound, I believe you can accept a DocumentModelList in the run method instead of a List of Doc refs which should simplify it.

I think it would also make sense for the BulkUpdateService to accept a DocumentModelList 
since this is what is returned by session.query (and could be very useful).

https://github.com/nuxeo/nuxeo/blob/8ce732309dc35ebe6c53aa8e0dd52757e8faf74a/modules/platform/nuxeo-automation/nuxeo-automation-features/src/main/java/org/nuxeo/ecm/automation/core/operations/collections/AddToCollectionOperation.java#L52

dy2288 (approved):

PR #747, opened 2021-04-06
FW-2336 FE: ErrorHandler v2

gmcauliffe:
#### Definition of Done
- [ ] Tests cover at least 85% of the functionality


dy2288 (changes requested):
Looks good to me!

If this is a code block we will be using again and again, consider making it a helper too (e.g. `NavigationHelpers.triggerError(history, error)`). That way the method can also be self-explanatory. We'll likely want to output some messages to the user for 500 or 400 errors, or directly log these in sentry - so another good reason to have a central method to trigger the error.

`      // Setting the errorStatusCode will trigger the render of the ErrorHandler component
       history.replace(history.location.pathname, {
         errorStatusCode: error?.response?.status,
       })`

gmcauliffe:
> Looks good to me!
> 
> If this is a code block we will be using again and again, consider making it a helper too (e.g. `NavigationHelpers.triggerError(history, error)`). That way the method can also be self-explanatory. We'll likely want to output some messages to the user for 500 or 400 errors, or directly log these in sentry - so another good reason to have a central method to trigger the error.
> 
> ` // Setting the errorStatusCode will trigger the render of the ErrorHandler component history.replace(history.location.pathname, { errorStatusCode: error?.response?.status, })`

I had the same thought and I did experiment with removing this into a helper method, see here: 
https://github.com/First-Peoples-Cultural-Council/fv-web-ui/pull/747/commits/cdca0d44123f3b1af7caeea4fab71b37a151c9b5
But ran into the "Only Call Hooks from React Functions" error, so I opted to stick to using it as is.

PR #1198, opened 2022-03-16
FW-3158 kids only content in songs/stories

dy2288:
+ Make kids only default checked


dy2288:
@gmcauliffe missed your note on FW-3158

PR #1364, opened 2022-08-08
Prepare for next development version

sarahfirstvoices:
- bumping all pom versions to 3.7.5-SNAPSHOT
- not changing any package.json versions because we are not currently tracking those


gmcauliffe (approved):
🥳

PR #1367, opened 2022-08-10
Fw 3457: part 1 of "related words" for character editing

sarahfirstvoices:
- added a new generic media browser for text items, which required some small changes to decouple document type from browser type
- use the new browser for editing "related words" on characters
- not in this review: other related assets for words


frontend/app_v2/src/components/Form/ArrayField.js
```diff
@@ -91,9 +91,7 @@ const AddDocButton = ({ addDoc, docType }) => {
       {/* Add Modal */}
       <Modal.Presentation isOpen={modalOpen} closeHandler={() => setModalOpen(false)}>
         <div className="w-1/2-screen mx-auto rounded-lg overflow-hidden">
-          {(docType === DOC_IMAGE || docType === DOC_VIDEO || docType === DOC_AUDIO) && (
-            <MediaBrowser.Container chooseMediaHandler={chooseDocHandler} docType={docType} />
-          )}
+          <MediaBrowser.Container chooseMediaHandler={chooseDocHandler} docType={docType} />
```
sarahfirstvoices, line 94: I removed this validation since it already listed in the MediaBrowser props and it makes one less place to update when we add new browser types, but I can put it back if folks disagree with this.

frontend/app_v2/src/components/MediaBrowser/MediaBrowserContainer.js
```diff
@@ -4,9 +4,17 @@ import PropTypes from 'prop-types'
...
+import { DOC_AUDIO, DOC_IMAGE, DOC_VIDEO, DOC_WORD } from 'common/constants'
 
 function MediaBrowserContainer({ chooseMediaHandler, docType }) {
+  const browserTypes = {}
+  browserTypes[DOC_AUDIO] = 'audio'
+  browserTypes[DOC_IMAGE] = 'images'
+  browserTypes[DOC_VIDEO] = 'video'
+  browserTypes[DOC_WORD] = 'list'
+
+  const browserType = browserTypes[docType]
```
sarahfirstvoices, line 16: There are a few little changes to separate the document type from the browser type, to allow a browser to be used for more than one document type.

frontend/app_v2/src/components/MediaBrowser/MediaBrowserData.js
```diff
@@ -82,7 +81,6 @@ function MediaBrowserData({ docType }) {
     loadRef,
     media: data,
     searchValue: searchInputValue,
-    type: pluralDocTypeLabel,
```
sarahfirstvoices, line 85: Since type is now only used to select the presentation style, I took it out of the data layer

frontend/app_v2/src/components/MediaBrowser/mediaDataAdaptor.js
```diff
@@ -47,6 +47,19 @@ const mediaDataAdaptor = ({ type, data }) => {
     return formattedData
   }
 
+  if (type === DOC_WORD) {
+    const { properties } = data
+    const formattedData = {
+      id: data?.uid,
+      filename: data?.title,
+      title: properties?.['dc:title'],
+      description: properties?.['dc:description'],
```
sarahfirstvoices, line 56: Not sure whether description is the best field to display for words. Definitions may be better?

sarahfirstvoices:
Fixing the duplication would require a bit of a refactor of the MediaBrowsers and I'd rather deal with it in a separate review, if folks are up for that.

PR #1376, opened 2022-08-23
fw-3644, updated docker image config for apple m2 machines

sarahfirstvoices:


Dockerfile-be
```diff
@@ -1,4 +1,8 @@
-FROM public.ecr.aws/a6r5f2i2/nx-base:latest
+# For Intel machines uncomment the following line:
+# FROM public.ecr.aws/a6r5f2i2/nx-base:latest
+
+# For Apple M2 machines uncomment the following line instead:
+FROM public.ecr.aws/a6r5f2i2/nx-base:arm64-aug19
```
sarahfirstvoices, line 5: I made this the default since it seems like the most likely scenario in the future. Open to other opinions on how to handle this

sarahfirstvoices: E.g., should we add this file to the .gitignore since we may each have different local configuration?

docker/README.md
```diff
@@ -17,15 +17,17 @@ This environment is setup for localhost work. It includes:
...
+1. Set two environment variables CYPRESS_FV_USERNAME and CYPRESS_FV_PASSWORD to any convenient username and password (e.g., your first name). These will be passed into the docker container and used to create an admin account during the initial setup. After setting the environment variables they can be checked by running the following in your terminal window: `printenv`
 
-3. Clone the fv-web-ui repository.
+1. Clone the fv-web-ui repository.
 
     ```
     git clone https://github.com/First-Peoples-Cultural-Council/fv-web-ui.git
     ```
+   
+1. Configure which Docker image to use for local development. Edit the Dockerfile-be file and uncomment either line 2 or line 5 depending on your machine's chip architecture.
```
sarahfirstvoices, line 29: This is the actual new instruction. Sorry for the noise-- I switched the numbers because it is annoying to try to keep them correct by hand

Dockerfile-be
sahil97: Maybe we can keep the default one and add the arm64 version in the comments ? Since I believe the rest of the team is using amd image.

sarahfirstvoices: Good call. Then nothing changes for other folks

sahil97 (approved):

PR #1387, opened 2022-08-31
Fw 3457 refactor mediabrowsers

sarahfirstvoices:
* reuse SearchSelector in MediaBrowser components
* moved main/aside elements from reusable components into page-level containers
* reuse ArrayBrowserField for media/category selectors
* renamed various ArrayFields


sarahfirstvoices:
Sonarbot failed this for duplication, but even sonarbot agrees that all the duplication would go away after merging. (It's all the files I split in two-- presumably the new files are being compared to the old files on the master branch.) Is there a way to ignore it?

gmcauliffe (approved):

PR #1398, opened 2022-09-07
Fw 3676 landingpage navbar

Cara-Barter:
#### Definition of Done
- navbar complete, needs proper arrow icon - will add with next ticket.  Background is transparent at top and charcoal once you start scrolling.


gmcauliffe (approved):
Off to the races! 🥇

PR #1401, opened 2022-09-07
Fw 3685 explore languages page initial setup

sahil97:
- Added Explore Language page
- Added site cards for showing different sites according to the site's visibility
- Added stylings for the explore language page and cards
- Added media queries for mobile layout support


frontend/app_v2/src/components/Languages/LanguagesFrame.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react'
+
+// FPCC
+import LanguagesContainer from './LanguagesContainer'
+import SectionTitle from '../SectionTitle/index'
```
gmcauliffe, line 5: Please avoid relative imports, instead import from `'components/SectionTitle'`, and then render using `<SectionTitle.Presentation ...`

gmcauliffe (changes requested):
Nice work, Sahil. If you could please go through and replace any relative imports with absolute imports using our webpack aliases, then you will be good to merge. Thanks!

gmcauliffe (approved):

PR #1404, opened 2022-09-08
fw-3581, improve backend test stability

sarahfirstvoices:
* prevent CoreSession errors during state transitions
* fix test data for media tests (merge problem)
* ignore unreliable tests in Publisher and Core Testing Framework modules


plasticviking (approved):
👌

PR #1406, opened 2022-09-09
FW-3686-explore-languages-page-adding-colors-to-parent-languages

sahil97:
- added a new config file containing colors for the parent languages
- added logic to fetch the rgb value of the parent colors from API


frontend/app_v2/src/components/Languages/LanguagesData.js
```diff
@@ -38,8 +52,19 @@ function LanguagesData() {
     return output
   }
 
+  const fetchColors = (parentLanguagesEntries) => {
```
gmcauliffe, line 55: NB: See the conversation thread re: DataAdaptors. The naming convention would apply to your fetchColors function.

gmcauliffe (approved):

PR #1410, opened 2022-09-13
hid placeholder on Landing Page, added image so it didn't look weird

Cara-Barter:
hid all placeholders on Landing Page.  Added image so that the footer wasn't crowding the NavBar


frontend/app_v2/src/components/LandingPage/LandingPagePresentation.js
```diff
@@ -52,7 +51,7 @@ function LandingPagePresentation({ data }) {
       <footer>
         <Footer.Presentation />
       </footer>
```
sarahfirstvoices, line 53: The first two footer links are broken on this. Is there an easy way to fix or hide that for this release? I don't think it's a deal breaker

Cara-Barter: weird.  It's that same footer component that is used on the rest of the site.

sarahfirstvoices: It looks like the urls are expecting a site value that is undefined on the landing page

Cara-Barter:
FW-3697 Hide Landing Page placeholder

sarahfirstvoices (approved):

PR #1426, opened 2022-09-27
FW-3640 add stepper to dictionarycrud

gmcauliffe:
#### Definition of Done


frontend/app_v2/src/components/DictionaryCrud/DictionaryCrudData.js
```diff
@@ -48,28 +49,28 @@ function DictionaryCrudData({ docType }) {
...
-      'fv:related_videos': data?.videos,
-      'fvaudience:children': data?.kidFriendly === 'true' ? true : false,
+      'dc:title': _formData?.title,
+      'fv:definitions': _formData?.translations,
+      'fv:related_audio': _formData?.audio,
+      'fv:related_assets': _formData?.relatedAssets,
+      'fv:related_pictures': _formData?.images,
+      'fv:related_videos': _formData?.videos,
+      'fvaudience:children': _formData?.kidFriendly === 'true' ? true : false,
       'fv:available_in_childrens_archive': null, // V1_FUDGE - clearing old property
```
sarahfirstvoices, line 61: I thought we decided to hide the v1 data but not delete it. This looks like it would set the old property to null for the update request.

frontend/app_v2/src/components/DictionaryCrud/DictionaryCrudPresentation.js
```diff
@@ -81,235 +81,252 @@ function DictionaryCrudPresentation({ backHandler, dataToEdit, submitHandler, do
...
+      setActiveStep(activeStep - 1)
+    } else {
+      backHandler()
     }
+  }
+
+  const onFinishClick = () => {
+    if (activeStep !== lastStep) stepHandle(lastStep)
     return
   }
```
sarahfirstvoices, line 112: I wonder if these step handling functions could get wrapped up in a little hook, along with the activeStep state.

gmcauliffe: It could if it was going to be reused and was less specific to this particular form.

frontend/app_v2/src/components/DictionaryCrud/DictionaryCrudData.js
gmcauliffe: Yes, that's true. Shall I ignore the existing data then at this stage?

frontend/app_v2/src/components/Form/Stepper.js
```diff
@@ -0,0 +1,85 @@
...
+                    <span className="text-gray-500 group-hover:text-gray-900">{`0${stepIdx + 1}`}</span>
+                  </span>
+                  <span className="ml-4 text-sm font-medium text-gray-500 group-hover:text-gray-900">{step.title}</span>
+                </span>
+              </button>
+            )}
+
+            {stepIdx !== steps.length - 1 ? (
+              <>
+                {/* Arrow separator for lg screens and up */}
```
sahil97, line 52: Should we separate the SVGs into their own components such as icons to keep them separate or if they're only used once, can we put them in the file ?

frontend/app_v2/src/components/DictionaryCrud/DictionaryCrudData.js
gmcauliffe: After discussing came to the conclusion to address this in a new ticket: [FW-3727](https://firstvoices.atlassian.net/browse/FW-3727)

sarahfirstvoices (approved):

frontend/app_v2/src/components/Form/Stepper.js
gmcauliffe: I just forgot to separate this one out, but you ask a good question. For now I will move this SVG into it's own file for consistency, but I will put it to the group to consider alternative approaches for SVGs. Thanks!

PR #1440, opened 2022-10-03
Fw 3579 only superadmin should be able to create or edit keyboard app no settings widgets

sahil97:
#### Definition of Done


frontend/app_v2/src/components/WidgetCrud/WidgetCrudData.js
```diff
@@ -21,31 +21,35 @@ import {
...
+    isSuperAdmin && WIDGET_KEYBOARDS,
+    WIDGET_LOGO,
+    isSuperAdmin && WIDGET_STATS,
+    WIDGET_TEXT,
+    WIDGET_TEXTCONCISE,
+    WIDGET_TEXTFULL,
+    //   WIDGET_TEXTMULTI,
+    isSuperAdmin && WIDGET_WOTD,
+  ].filter(Boolean)
+
```
sarahfirstvoices, line 52: All of these solutions look great, but I notice we are maintaining the list of SuperAdmin-only widgets in three different places. My limit for repetition is usually two. Could we put the list somewhere external and reuse it instead? Maybe a little useWidgetAccess hook or something? We could think about what would make the most sense to reuse-- maybe a function like hasWidgetAccess(user), or maybe just a list of superadmin-only widgets

sarahfirstvoices (changes requested):

frontend/app_v2/src/components/WidgetCrud/WidgetCrudData.js
sahil97: Good point. I like the idea of the hasWidgetAccess which can filter according the a flag supplied, e.g. isSuperAdmin which we have in the data layer of each component.
The function can be placed in the common folder which contain other helper files as well.
What do you think ?

sarahfirstvoices: Sounds great to me

sarahfirstvoices: This looks great! This is kind of a nitpick, but to make it more React-style, could you change widgetAccessHelpers into a hook (similar to useUserStore or the others that start with `use`)? Call me if you want to talk it through

sarahfirstvoices: We chatted and Sahil made the great point that since there is no state this doesn't really benefit from being a hook. I retract my comment!

sarahfirstvoices (approved):

PR #1444, opened 2022-10-05
Fw 3579 only superadmin should be able to create or edit keyboard app no settings widgets

sahil97:
#### Definition of Done


sarahfirstvoices (approved):
Wow, that list is getting even more reuse. Great!

PR #1483, opened 2022-10-24
FW-3486 crud story pages

gmcauliffe:
#### Definition of Done


frontend/app_v2/src/common/stringHelpers.js
```diff
@@ -193,7 +194,7 @@ export const isUUID = (str) => {
   if (!str) {
     return false
   }
-  const isId = str.match(/^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i)
+  const isId = str.match(UID_REGEX)
```
sarahfirstvoices, line 197: So technically I think this is a UUID with two u's. I think we use a bit of a mix of uuid and uid (as does probably everyone because it is hard to keep straight), but you could consider whether to use two u's here. Don't try to read this out loud ha ha.

frontend/app_v2/src/common/useValidationSchema.js
```diff
@@ -0,0 +1,62 @@
...
+    wysiwyg: (charCount) => {
+      return yup
+        .object()
+        .test('draftjs-length', `Maximum length for this field is ${charCount} characters`, (value) => {
+          const plainText = value?.getCurrentContent ? value.getCurrentContent().getPlainText(' ') : ''
+          return plainText?.length < charCount
+        })
+        .nullable()
+    },
+  }
```
sarahfirstvoices, line 32: I love having these reusable definitions. +1 for this hook.

frontend/app_v2/src/common/useValidationSchema.js
```diff
@@ -0,0 +1,62 @@
...
+    audio: compoundDefs.idArray,
+    images: compoundDefs.idArray,
+    videos: compoundDefs.idArray,
+  })
+
+  return {
+    storyCoverResolver: yupResolver(storyCoverValidator),
+    storyPageResolver: yupResolver(storyPageValidator),
+  }
+}
```
sarahfirstvoices, line 60: Are the specific schemas likely to be used in multiple places? If each schema is only used within one component it might be more maintainable to define the schema right in each component, maybe where we define the form that uses it.

frontend/app_v2/src/components/WidgetCrud/WidgetFormTextFull.js
```diff
@@ -15,7 +15,7 @@ function WidgetFormText({ cancelHandler, dataToEdit, submitHandler }) {
     textWithFormatting: yup
       .object()
       .test('draftjs-length', 'Maximum length for this field is 4500 characters', (value) => {
-        const plainText = value.getCurrentContent().getPlainText(' ')
+        const plainText = value?.getCurrentContent ? value.getCurrentContent().getPlainText(' ') : ''
```
sarahfirstvoices, line 18: This line seems like a candidate for a reusable function. Something like getTextValue(value, defaultValue) ?

frontend/app_v2/src/services/api/document.js
```diff
@@ -43,7 +43,7 @@ const document = {
     const body = {
       'entity-type': 'document',
       uid: id,
-      properties: properties,
+      properties: { ...properties, 'fv:modifiedv2': true },
     }
```
sarahfirstvoices, line 47: It worked! It's exciting to see this in a simple, central place.

frontend/app_v2/src/components/StoryCoverCrud/StoryCoverCrudContainer.js
```diff
@@ -5,21 +5,22 @@ import PropTypes from 'prop-types'
...
-  const { backHandler, dataToEdit, isCreate, submitHandler } = StoryCoverCrudData()
+function StoryCoverCrudContainer({ onNextClickCallback }) {
+  const { backHandler, dataToEdit, isCreate, submitHandler } = StoryCoverCrudData({ onNextClickCallback })
   return (
     <StoryCoverCrudPresentation
       backHandler={backHandler}
       isCreate={isCreate}
       dataToEdit={dataToEdit}
+      onNextClickCallback={onNextClickCallback}
       submitHandler={submitHandler}
```
sarahfirstvoices, line 16: Do we have any naming conventions about onSomeEvent vs someEventHandler vs handleSomeEvent type names?

frontend/app_v2/src/common/useValidationSchema.js
gmcauliffe: It was the individual schema definitions that I was trying to reduce repeating. By putting all of the schemas in one hook the individual definitions can be reused for multiple schema e.g. yup
```
      yup.object()
      .test('draftjs-length', 'Maximum length for this field is 4500 characters', (value) => {
        const plainText = value?.getCurrentContent ? value.getCurrentContent().getPlainText(' ') : ''
        return plainText?.length < 4500
      }),
```

frontend/app_v2/src/components/WidgetCrud/WidgetFormTextFull.js
gmcauliffe: See `useValidationSchema` comment.

frontend/app_v2/src/components/StoryCoverCrud/StoryCoverCrudContainer.js
gmcauliffe: There are some general conventions that I try to follow, but they could be further refined. Let's discuss :)

sarahfirstvoices (approved):

PR #1490, opened 2022-10-26
Added audio upload support

sahil97:
#### Definition of Done


frontend/app_v2/src/common/constants/docTypes.js
```diff
@@ -13,6 +13,7 @@ export const DOC_IMAGE = 'FVPicture'
 export const DOC_VIDEO = 'FVVideo'
 export const SUPPORTED_IMAGE_EXTENSIONS = ['jpeg', 'png', 'gif', 'tiff']
```
sarahfirstvoices, line 14: I know this isn't part of this review, but does uppy handle variations on these file extensions, like jpg and tif?

sahil97: Good point. I can check

sarahfirstvoices (changes requested):
Best solution is probably to switch from file extensions to mime types

frontend/app_v2/src/common/constants/docTypes.js
sahil97: Added extensions manually instead of the mime type to be more specific.

sarahfirstvoices (approved):

PR #1505, opened 2022-11-02
Fw 3781 crud for fv person

sahil97:
#### Definition of Done


modules/common/firstvoices-core-tests/src/main/java/ca/firstvoices/tests/mocks/services/MockDialectServiceImpl.java
```diff
@@ -431,13 +439,13 @@ private void generateMedia(CoreSession session, String path, DocumentModelList c
 
         // Alternate 1 and 2 speakers
```
sarahfirstvoices, line 440: Hmm, these media properties might need rethinking. The version in master matched the v1 data, but the changes needed for v2 are more complicated than just swapping contributors for persons. I think all the new fields have been implemented at the schema level now, so you could make the mock data generator match the v2 format. Details are on the v2 field mapping wiki page: https://firstvoices.atlassian.net/wiki/spaces/FIR/pages/79691807/V2+Data+Migration+Field+Mapping#FVAudio-FVImage-&-FVVideo

modules/common/firstvoices-data/src/main/resources/data/schemas/fvdictionary.xsd
```diff
@@ -42,6 +42,15 @@
...
   </xs:element>
+  <xs:element name="persons">
+    <xs:simpleType>
+      <xs:list>
+        <xs:simpleType>
+          <xs:restriction base="xs:string" ref:resolver="documentResolver" ref:store="id"/>
+        </xs:simpleType>
+      </xs:list>
+    </xs:simpleType>
+  </xs:element>
```
sarahfirstvoices, line 53: I don't think this is needed. The only place we currently plan to use FVPerson is for the speaker property of FVAudio documents. See the FVContributor migration plan for details: https://firstvoices.atlassian.net/wiki/spaces/FIR/pages/61341697/FVContributor+migration+plan#Fields-to-migrate

modules/common/firstvoices-publisher/src/main/java/ca/firstvoices/publisher/services/FirstVoicesSyncStateServiceImpl.java
```diff
@@ -55,7 +55,7 @@ public boolean isDependencyContainer(DocumentModel mainDoc, DocumentModel depend
         return true;
       }
 
-      String[] parents = new String[]{FV_RESOURCES_NAME, FV_CATEGORIES_NAME, FV_PHRASE_BOOKS_NAME, FV_CONTRIBUTORS_NAME};
+      String[] parents = new String[]{FV_RESOURCES_NAME, FV_CATEGORIES_NAME, FV_PHRASE_BOOKS_NAME, FV_PERSONS_NAME};
```
sarahfirstvoices, line 58: What does this do? Are we sure it isn't still used in v1? I wouldn't want to mess up the contributor data that is still coming in from v1

modules/security/firstvoices-security-user-registration/src/main/resources/OSGI-INF/extensions/ca.firstvoices.usermapping.xml
```diff
@@ -16,6 +16,7 @@
         userAttributes.put("referer", userObject.comments);
         userAttributes.put("created", userObject.comments);
         userAttributes.put("contributors", userObject.contributors);
+        userAttributes.put("persons", userObject.persons);
```
sarahfirstvoices, line 19: Are you sure this is the same "contributors" as FVContributors? I'm not sure we should be having to edit anything about users for this.

modules/security/firstvoices-security-user-registration/src/main/resources/schemas/fvuser.xsd
```diff
@@ -16,7 +16,10 @@
   <!-- inverse reference -->
   <xs:element name="groups" type="nxs:stringList"/>
 
+  <!-- V1_Fudge  -->
   <xs:element name="contributors" type="xs:string"/>
+
```
sarahfirstvoices, line 21: Same comment as above-- are we sure fvuser:contributors is related to this change?

sarahfirstvoices (changes requested):

modules/common/firstvoices-publisher/src/main/java/ca/firstvoices/publisher/services/FirstVoicesSyncStateServiceImpl.java
sahil97: So from the comments, I believe this service ensures when we are transitioning any object to a different state, all these linked objects are also transitioned to that state.
You are right that this should not impact Persons ( and I guess should not included contributors ? as their state needs to be managed independently ?)
Reverting this back.

modules/common/firstvoices-core-tests/src/main/java/ca/firstvoices/tests/mocks/services/MockDialectServiceImpl.java
sarahfirstvoices: Setting e.g., the image recorder to a person is still not going to give very good results, I think. It will be broken in both v1 and v2. I think we should make all the media data match the v2 format. So instead of including any fvm:recorder or fvm:source properties, the audio should have fvaudio:speakers and all 3 types should have fvm:acknowledgement

sarahfirstvoices (changes requested):

sarahfirstvoices (approved):

PR #1516, opened 2022-11-09
Fw 3817 validate speaker ids for media uploads

sahil97:
#### Definition of Done


modules/api/firstvoices-rest-api/src/main/java/ca/firstvoices/rest/helpers/DocumentUtils.java
```diff
@@ -0,0 +1,92 @@
...
+    DocumentModel speakerDialect = getValidatedParentDocument(context, (String) speaker.getPropertyValue("fva:dialect"));
+    String speakerDialectId = speakerDialect.getId();
+    String parentDocId = parentDoc.getId();
+    if(!speakerDialectId.equals(parentDocId)) {
+      throw new IllegalArgumentException();
+    }
+    return true;
+  }
+
+  public static DocumentModel getValidatedParentDocument(WebContext context, String parentDocumentId) throws DocumentNotFoundException {
```
sarahfirstvoices, line 50: I don't think the "parent document" language makes sense outside of the MediaUploadObject. If this is going to be a reusable public util function, I think it should just be called getValidatedDocument, and then the error message will need to be handled in MediaUploadObject

modules/api/firstvoices-rest-api/src/main/java/ca/firstvoices/rest/helpers/DocumentUtils.java
```diff
@@ -0,0 +1,92 @@
...
+  public static boolean validateSpeakerId(WebContext context, String speakerId, DocumentModel parentDoc) throws IllegalArgumentException {
+    final CoreSession session = context.getCoreSession();
+    String verifiedSpeakerId = validateDocumentType(context, speakerId, FV_PERSON);
+    DocumentModel speaker = session.getDocument(new IdRef(verifiedSpeakerId));
+    DocumentModel speakerDialect = getValidatedParentDocument(context, (String) speaker.getPropertyValue("fva:dialect"));
+    String speakerDialectId = speakerDialect.getId();
+    String parentDocId = parentDoc.getId();
+    if(!speakerDialectId.equals(parentDocId)) {
+      throw new IllegalArgumentException();
+    }
```
sarahfirstvoices, line 46: This bit of checking would be more clear if it was extracted to a function called something like hasDialect(documentId, dialectId) or validateDocumentHasDialect. I think you could also add a specific message for the exception, something like "Document [123] does not have the expected parent [456]."

modules/api/firstvoices-rest-api/src/main/java/ca/firstvoices/rest/helpers/DocumentUtils.java
```diff
@@ -0,0 +1,92 @@
...
+
+import java.net.URL;
+import org.apache.logging.log4j.util.Strings;
+import org.nuxeo.ecm.core.api.CoreSession;
+import org.nuxeo.ecm.core.api.DocumentModel;
+import org.nuxeo.ecm.core.api.DocumentNotFoundException;
+import org.nuxeo.ecm.core.api.IdRef;
+import org.nuxeo.ecm.webengine.model.WebContext;
+
+public class DocumentUtils {
```
sarahfirstvoices, line 16: Please add unit tests for this class.

modules/api/firstvoices-rest-api/src/main/java/ca/firstvoices/rest/helpers/DocumentUtils.java
```diff
@@ -0,0 +1,92 @@
...
+      case "mp3":
+        return FV_AUDIO;
+      case "mov":
+      case "mp4":
+        return FV_VIDEO;
+      default:
+        String errorMsg = "Unsupported file type: [" + fileType + "]. Supported types are: jpg, gif, png, tif, wav, mp3, mov, and mp4.";
+        throw new IllegalArgumentException(errorMsg);
+    }
+  }
```
sarahfirstvoices, line 90: Now that these functions are all collected into one class, I notice they don't all have consistent names and behaviour. This will make them harder to use and harder to maintain. Let's chat about how to standardize them.

Specific questions:
- can functions with matching names have matching behaviour? E.g., getValidatedX means it returns a value or throws an exception, while validateX is void but may throw an exception.
- does it make sense to generate all these exceptions in a utility class, or should they be handled by the consumers?

modules/api/firstvoices-rest-api/src/main/java/ca/firstvoices/rest/helpers/DocumentUtils.java
```diff
@@ -0,0 +1,92 @@
...
+    try {
+      final CoreSession session = context.getCoreSession();
+      return session.getDocument(new IdRef(parentDocumentId));
+    }
+    catch(DocumentNotFoundException e) {
+      throw new IllegalArgumentException("Valid parent document id is required", e);
+    }
+  }
+
+  public static URL getValidatedUrl(String urlString) {
```
sarahfirstvoices, line 60: This function has nothing to do with documents and doesn't really belong here. Is it being reused, or could it go back where it came from?

modules/api/firstvoices-rest-api/src/main/java/ca/firstvoices/rest/helpers/DocumentUtils.java
```diff
@@ -0,0 +1,92 @@
...
+        throw new IllegalArgumentException();
+      }
+    }
+    catch (Exception e) {
+      throw new IllegalArgumentException(String.format("Not a valid %s: [%s]", type, docId), e);
+    }
+    return docId;
+  }
+
+  public static boolean validateSpeakerId(WebContext context, String speakerId, DocumentModel parentDoc) throws IllegalArgumentException {
```
sarahfirstvoices, line 37: Not sure this function belongs here-- it's more specific than "DocumentUtils" would suggest, and we are only using it in one place. Should it be in MediaUploadObject instead?

modules/api/firstvoices-rest-api/src/main/java/org/nuxeo/ecm/restapi/server/jaxrs/firstvoices/AbstractSearchlikeObject.java
```diff
@@ -80,24 +79,6 @@ protected SearchResults.SearchDomain resolveDomainFromQueryString(String domain)
     }
   }
 
-  protected String validateDocumentType(String docId, String type) throws IllegalArgumentException {
```
sarahfirstvoices, line 83: You could consider keeping this function as a little wrapper that passes the session to the utility function.

sarahfirstvoices (changes requested):

modules/api/firstvoices-rest-api/src/main/java/ca/firstvoices/rest/helpers/DocumentUtils.java
sahil97: I believe it can be like you mentioned in your first point, but the validateX also returns true if file is of valid type, instead of boolean. Then we can also use it in conditional clauses.
What do you think ?
I'm not sure on the exceptions part, should it be like the utility classes generate exceptions and the consumer handles them or adds messages and then forward them to their parent ?

sarahfirstvoices: Boolean instead of void for the validateX methods sounds good to me.

And I think since all these methods have validate in the name and are intended to check values we expect to be correct, it's appropriate to throw exceptions. Not sure IllegalArgumentException is quite right, but it's good enough

modules/api/firstvoices-rest-api/src/main/java/org/nuxeo/ecm/restapi/server/jaxrs/firstvoices/MediaUploadObject.java
```diff
@@ -132,6 +137,17 @@ public Response markComplete(MediaUploadCompleteRequest requestBody) throws Unsu
       return response(Status.INTERNAL_SERVER_ERROR.getStatusCode(), new NuxeoException("Could not retrieve media object [" + getObjectKey(url) + "]" + e.getMessage()));
     }
 
+    if(mediaDocType.equals(FV_AUDIO)){
```
sarahfirstvoices, line 140: This could go even earlier. There's no need to even download the file if the parameters are invalid, so you can move this validation up to line 119-ish, as soon as you know the media type.

modules/api/firstvoices-rest-api/src/test/java/org/nuxeo/ecm/restapi/server/jaxrs/firstvoices/DocumentUtilsTest.java
```diff
@@ -0,0 +1,191 @@
...
+  protected CoreSession session;
+
+  @Test
+  public void validateDocumentTypeTest_dialect() {
+    String sampleDialectId = dataCreator.getReference(session, "testArchive").getId();
+    Boolean isDialect = validateDocumentType(session, sampleDialectId, FV_DIALECT);
+    assertTrue(isDialect);
+  }
+
+  @Test
```
sarahfirstvoices, line 55: This is fine the way it is, but to save effort in the future there's no need to test multiple document types here, since they don't exercise any different lines of code. One success response and one of each type of failure would be enough

modules/api/firstvoices-rest-api/src/test/java/org/nuxeo/ecm/restapi/server/jaxrs/firstvoices/DocumentUtilsTest.java
```diff
@@ -0,0 +1,191 @@
...
+  public void validateDocumentTypeTest_character() {
+    String sampleCharId = dataCreator.getReference(session, "minChar1").getId();
+    Boolean isChar = validateDocumentType(session, sampleCharId, FV_CHARACTER);
+    assertTrue(isChar);
+  }
+
+  @Test
+  public void validateDocumentTypeTest_wrongObjectType() {
```
sarahfirstvoices, line 105: Another type of failure we should test here is if the document id is not valid (document not found)

modules/api/firstvoices-rest-api/src/test/java/org/nuxeo/ecm/restapi/server/jaxrs/firstvoices/MediaUploadSpeakerValidationTest.java
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  @Test
+  public void testValidateDocumentHasDialect_invalidParentDocumentId() {
+    try {
+      Boolean isValid = mediaUploadObject.validateDocumentHasDialect("speaker-id", "not-parent-id");
+    }
+    catch (IllegalArgumentException e) {
+      assertTrue(e.getMessage().contains("Document [speaker-id] does not have the expected parent [not-parent-id]"));
+    }
```
sarahfirstvoices, line 69: If this doesn't throw an error does the test fail? I think it needs a failing assertion after the catch block, just in case

sarahfirstvoices: Same for the other error tests below

modules/api/firstvoices-rest-api/src/main/java/org/nuxeo/ecm/restapi/server/jaxrs/firstvoices/MediaUploadObject.java
```diff
@@ -286,14 +246,30 @@ private void deleteS3Object(URL url) throws UnsupportedEncodingException {
...
+      try {
+        String[] speakerIds = request.getSpeaker();
+        for (int i = 0; i < speakerIds.length; i++) {
+          if (!validateSpeakerId(getContext(), speakerIds[i], parentDoc)){
+            throw new IllegalArgumentException();
+          }
+        }
+        mediaDoc.setPropertyValue("fvaudio:speakers", speakerIds);
+      } catch (Exception e){
+        String errorMsg = "Invalid speaker Id(s): Please verify Speaker Ids and the associated Dialect";
```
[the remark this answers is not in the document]
sarahfirstvoices: Looks great, thanks for these edits

modules/api/firstvoices-rest-api/src/test/java/org/nuxeo/ecm/restapi/server/jaxrs/firstvoices/MediaUploadSpeakerValidationTest.java
sahil97: Good catch. Didn't think of the passing case.

sarahfirstvoices (approved):

PR #1541, opened 2022-11-29
fw-3533, new SearchTypeSelector and custom table headers for search results

sarahfirstvoices:
Summary of changes:
- Tabs component value is the searchType string instead of a tab object now
- relevant pages all use new SearchTypeSelector, which is just Tabs preloaded with the word/phrase/both data
- table headers and "no results for x" messages use search type labels
- some small behaviour fixes for ui combo moves like filtering by word/phrase and then clicking "reset search"


frontend/app_v2/src/components/SearchSelector/index.js
```diff
@@ -0,0 +1,5 @@
+import SearchSelectorPresentation from 'components/SearchSelector/SearchSelectorPresentation'
+
+export default {
+  Presentation: SearchSelectorPresentation,
```
sarahfirstvoices, line 4: I was checking the index.js files for the new search-related components and realized this old one was missing as well

gmcauliffe (approved):

PR #1543, opened 2022-11-30
FW-3857 visibility selector part 1

gmcauliffe:
Created Visibility select component that 
- takes a uuid and an initial doc state and fires a request to change the visibility of the document
- Includes confirmation dialog
- checks visibility of the parent site and offers visibility options accordingly

Added Visibility select component to the header of the Custom pages edit page


frontend/app_v2/src/components/DictionaryCrud/DictionaryCrudData.js
```diff
@@ -107,11 +107,13 @@ function DictionaryCrudData({ docType }) {
     })
     if (response?.uid) {
       if (formData?.visibility !== dataToEdit?.visibility) {
-        api.visibility.update({ id: entryId, newVisibility: formData?.visibility })
+        await api.visibility.update({ id: entryId, newVisibility: formData?.visibility })
```
gmcauliffe, line 110: Was redirecting before response from visibility update endpoint was received.

frontend/app_v2/src/components/DictionaryCrud/DictionaryCrudData.js
```diff
@@ -107,11 +107,13 @@ function DictionaryCrudData({ docType }) {
...
       }
       setNotification({ type: 'SUCCESS', message: 'Success! Your dictionary entry has been saved.' })
       queryClient.invalidateQueries([docType, entryId])
-      backHandler()
+      setTimeout(function () {
+        backHandler()
+      }, 1000)
```
sarahfirstvoices, line 116: Is there a specific reason to have a 1 second timer? I wonder if this could cause problems on slow connections. Would it make sense instead to use success and error callbacks and let the request time out on its own?

sarahfirstvoices (approved):

PR #1544, opened 2022-11-30
fw-3533, more search result table headers

sarahfirstvoices:
Contextual table labels for dictionary and search pages. (Missed them before!)


Cara-Barter (approved):
LGTM

PR #1570, opened 2022-12-08
Fw 3850 adapt text widget

Cara-Barter:
adapted Text widget form for new fields and fixed image in Icons widget


gmcauliffe (approved):
Just some small changes as discussed. LGTM 👍

PR #1575, opened 2022-12-15
fw-3794, expandable story page previews

sarahfirstvoices:
These are working, but the edit forms themselves have not been overhauled to match the ux designs yet. (See FW-3900)


frontend/app_v2/src/common/useWysiwygState.js
```diff
@@ -1,20 +1,23 @@
...
-import { isStringWithLength, extractTextFromHtml } from 'common/stringHelpers'
+import { isStringWithLength, extractTextFromHtml, safeJsonParse } from 'common/stringHelpers'
 
 function useWysiwygState() {
   // Converts an HTML string into Draft-js ContentState
   const getWysiwygStateFromHtml = (htmlString) => {
-    if (isStringWithLength(htmlString)) {
-      const blocksFromHTML = convertFromHTML(htmlString)
-      return ContentState.createFromBlockArray(blocksFromHTML.contentBlocks, blocksFromHTML.entityMap)
+    const textFromHtml = extractTextFromHtml(htmlString)
```
gmcauliffe, line 12: Using `extractTextFromHtml` will strip the formatting from the data. Is there a way that we can acheive what you want here without losing the formatting from v1 data?

PR #1604, opened 2023-01-12
Fw 3956 v 2 be return speakers details in the context parameters for media files

sahil97:
#### Definition of Done


modules/common/firstvoices-enrichers/src/main/java/ca/firstvoices/enrichers/MediaEnricher.java
```diff
@@ -99,6 +99,20 @@ private ObjectNode constructMediaJSON(DocumentModel doc) {
         jsonObj.set("recorders", recorderArray);
       }
 
+      // Process "fvaudio:speaker" values
+      String[] speakerIds = (String[]) doc.getProperty("fvaudio", "speakers");
+      if (speakerIds != null) {
```
sarahfirstvoices, line 104: It would probably help the frontend if we returned an empty array for the null case

sarahfirstvoices (approved):

PR #1637, opened 2023-02-02
fw-3994, use same constants for group names everywhere

sarahfirstvoices:
- Eliminated the duplicate constants from security-user-registration module and used the ones from security module everywhere
- Also renamed some visibility constants to avoid confusion between MEMBERS_GROUP and MEMBERS_VISIBILITY


NolanVH (approved):
Having constants named differently in various location was previously very confusing so this is a great change! Thanks!

Changes build successfully on a local machine. All changes look good to me.

PR #1640, opened 2023-02-03
fw-3994, change Public Members permitted folders

sarahfirstvoices:
- Remove top-level read access for the public members group, and instead add read access to specific dialect folders when they are published. This eliminates a rare case where a Public Member could access an Enabled site without being a member of the site. 
- Add read access for the Members group when a dialect is published. This allows logged-in members of any site to see all public sites.
- This does not change any group folder permissions on existing installations, only on newly published dialects.


modules/security/firstvoices-security-user-registration/src/main/java/ca/firstvoices/operations/FVRequestToJoinDialectAdminAction.java
```diff
@@ -68,6 +68,7 @@ public class FVRequestToJoinDialectAdminAction {
   public static final String ID = "LanguageAdmin.RequestToJoinDialectAction";
 
   private static final String GROUP_KEY = "group";
+  private static final String MEMBERS_KEY = "members";
   private final List<String> validStartingStatuses = Arrays.asList("NEW", "IGNORE", "ACCEPT");
```
sarahfirstvoices, line 72: Unrelated cleanup to clarify what this string actually means

modules/security/firstvoices-security-user-registration/src/main/java/ca/firstvoices/operations/FVRequestToJoinDialectAdminAction.java
```diff
@@ -268,19 +269,19 @@ private boolean isRelevantACE(String group, ACE ace) {
       }
 
       switch (ace.getPermission()) {
-        case org.nuxeo.ecm.core.api.security.SecurityConstants.READ:
+        case CustomSecurityConstants.READ:
           return CustomSecurityConstants.MEMBERS_GROUP.equals(group)
```
sarahfirstvoices, line 273: Unrelated cleanup to use our custom constants everywhere

sahil97 (approved):

sarahfirstvoices:
Closing this due to our decision to replace the backend rather than keep fixing it.

PR #1669, opened 2023-02-23
Fw 3900 story editing forms ux

Cara-Barter:
story crud matches ux designs


Cara-Barter:
Styling changes are not consistent with the rest of the site.  We will need to review the UX process.

PR #1684, opened 2023-03-07
FW-4097 Media Browser Single Selector

NolanVH:
This change updates the media browser single selector green checkmark to update to the last clicked item and deselect any other selected items. Previously, a user was required to unselect any selected item before gaining the ability to select a new item.


frontend/app_v2/src/components/MediaCrud/MediaCrudData.js
```diff
@@ -85,7 +85,7 @@ function MediaCrudData({ docType, maxFiles }) {
     if (selectedMedia.some((elemId) => elemId == docId)) {
       updatedSelectedMedia = updatedSelectedMedia.filter((elemId) => elemId != docId)
     } else if (maxFiles && updatedSelectedMedia.length == maxFiles) {
-      return
+      updatedSelectedMedia = [docId]
```
sarahfirstvoices, line 88: When this is used to select multiple media files, and the user has already selected the max amount, it looks like clicking another file might remove all the previous selections. Is that true or am I missing something?

NolanVH: This is a good point, and you are correct. Currently, the maxFiles value is only ever set to 1 or nonexistent, so this is not a problem, but could become one later. 

I have updated the functionality to only replace the selector if maxFiles equals 1. If maxFiles is greater than 1 then the user will be able to select up to the maxFiles value and then they will have to deselect an item before adding another.

Good catch, thanks @sarahfirstvoices!

sarahfirstvoices (approved):

PR #1691, opened 2023-03-08
FW-4135-link-tabs

Cara-Barter:
internal links open in the same tab, external in a new tab


NolanVH (approved):
All changes look good to me. Thanks, @Cara-Barter!

PR #1692, opened 2023-03-09
FW-4109 Change Visibility Selector Default Values

bronson-c:
#### Definition of Done

This change updates the Listbox selector so that if there is no selected value, the default value is the last possible option if there are options available for selection. This ensures that the default values on pages with visibility selector work as specified in the ticket (New -> Team Only, Enabled -> Members Only)


bronson-c:
This change could also be handled without touching the Listbox component, but it would require changes to every form (All widgets, dictionary entries, custom pages etc.). Changing the listbox handles the issue everywhere.

gmcauliffe (approved):
The detailed comments you added really helped to add context and make your reasoning clear.  Thanks 👍

PR #1697, opened 2023-03-15
FW-4100-Story-Crud-Preview-Image

Cara-Barter:
added conditional statement for when no cover image selected


gmcauliffe (approved):
Looks good. Have you tested what happens if the coverVisual is a video?

